PR #2, opened 2016-07-06
Comp2016 - Firmware

jgkamat:
The firmware PR for comp2016. For more info, see:
https://github.com/RoboJackets/robocup-software/pull/767


guyfleeman:
How close is this? I don't want to do too much with the build system until this is good to go.

jgkamat:
build system has already been seperated with only minor changes to this branch. Once this is merged, we'll use the common submodule stuff.

guyfleeman:
I'm getting a cmake error. Also googletest is still in submodules, do we really need it?

jgkamat:
Can you post the error? 
also run: `git submodule update --init`

I think we still need googletest for firmware tests.

guyfleeman:
We do firmware tests?

jgkamat:
I think so?

try `make test-firmware` if my memory is right...

guyfleeman:
I'm still getting CMake errors but I'd rather address those once the split is finalized. It is a target though. I don't understand how we unit test anything in firmware.

justbuchanan:
The firmware tests are very minimal and not particularly helpful.  The goal was to test some of the firmware logic, not the actual hardware interfaces, but it turned out to be pretty tough to separate the two.

guyfleeman:
Is it worth it even trying? I understand it in firmware situations where more logic is being done, but we basically do the minimum that's required to translate radio packets to outputs right?

jjones646:
> but it turned out to be pretty tough to separate the two

^ this. It's hard to separate since hardware affects the firmware in runtime (ie. no radio board or whatever else could go wrong). Making sure the firmware targets compile is about all you can do without running it on hardware.

justbuchanan:
As @jjones646, compiling is about as good of a check as you can run on the computer.  If you're working with the mbed, you can also check out the programs in the `hw-test` directory which test some of the individual hardware components/drivers.

firmware/common2015/drivers/ball_sense/BallSense.hpp
```diff
@@ -36,7 +39,7 @@ class BallSense {
 
     // If the light reading and dark reading are closer than this value, the
     // beam is considered broken
-    const int sense_threshold = 5;
+    const int sense_threshold = 500;
```
guyfleeman: Anyone know if this is a good value or if finer tuning is required?

guyfleeman:
@jgkamat most of the tweaks and changes here actually look gtg but I'm sure complications will arise as we get deeper into it

jgkamat:
I'll look at this once software is merged (I haven't looked at it at all, and I'm not sure if I would be able to provide anything useful anyway), but that's great to hear!

firmware/common2015/drivers/ball_sense/BallSense.hpp
chachmu: Me and Justin tuned it in the shop and it seemed pretty good, we tested it in both dark and light conditions.

firmware/robot2015/src-ctrl/modules/ControllerTaskThread.cpp
```diff
@@ -78,41 +84,34 @@ void Task_Controller(void const* args) {
...
-
-    Pid motor2pid(0.0, 0.0, 0.0);
-
-    std::vector<int16_t> duty_cycles;
+    array<int16_t, 5> duty_cycles{};
 
-    const uint16_t kduty_cycle = 0;
-    duty_cycles.assign(5, kduty_cycle);
+    // pidController.setPidValues(1.5, 0.05, 0);  // TODO: tune pid values
+    pidController.setPidValues(0.8, 0.05, 0);
```
jgkamat: Are these good?

jgkamat:
LGTM as well, dosen't seem like there were too many changes.

jgkamat:
Is this good to merge? I want to merge this by next weekend if possible.

@guyfleeman

jgkamat:
Crossing figners...

PR #22, opened 2016-08-31
Removed AUR package for ARM cross compilation.

jpfeltracco:
I think we had some issues in the past with the arm compiler in the community repos so we were still using one from the AUR which really isn't ideal. The community versions seem to be fixed up and compiling fine now.

### DIFF
--- a/util/arch-packages-aur.txt
+++ b/util/arch-packages-aur.txt
@@ -1,8 +1,5 @@
 # This is a list of packages that we need for Arch Linux that
 # we'll install from the AUR using yaourt
 
-# mbed compiler suite
-gcc-arm-none-eabi-bin
-
 # optional - needed if you want to use the SpaceNavigator 3d mouse to drive robots
 #spacenavd
--- a/util/arch-packages.txt
+++ b/util/arch-packages.txt
@@ -50,6 +50,11 @@ libspnav
 # joystick library
 sdl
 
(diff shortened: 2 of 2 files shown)


jgkamat:
Ooh, I never noticed they got put into community, thats nice! I'll merge this sometime this weekend (I want to test it out a bit)

jpfeltracco:
Yeah, I think Justin had moved us over to the community package a while back and it was missing some function in a header we were using, but it seems to be all fixed up now, I'll be testing with the new package too and let you know if anything breaks!

jgkamat:
Can you add [arm-none-eabi-gcc](https://www.archlinux.org/packages/community/x86_64/arm-none-eabi-gdb/) to the package list? I think it's needed for the debugging mode that may or may not work...
I don't know too much about it, but I think you can run it with `make robot2015-gdb`

jpfeltracco:
I think gcc is already on there, do you mean gdb?

jgkamat:
Whoops, my bad, yah I meant gdb. The link to the package should be right though...

jpfeltracco:
Alrighty, added arm-none-eabi-gdb in!

jgkamat:
Nice, I'll go ahead an merge it now, it seems good to me =)

Hopefully the transition won't cause any issues, but I don't think it will.

PR #23, opened 2016-09-05
Added existing files from avr-spi

jgkamat:
Imported from RoboJackets/robocup-software#692

Sorry guys but there were just too many merge conflicts to do this any other way than squash everything into a single commit (and make sure the two diffs are the same).

Good luck...

PR #24, opened 2016-09-09
Kickerboard firmware

jpfeltracco:
This is to replace #23, which was imported from RoboJackets/robocup-software#692.

I rebased on top of master and fixed up conflicts with master, I'm making a new PR because I think the branch name "avr-spi" is really bad, and I wanted to keep a non-rebased copy of avr-spi around for reference until this is merged in.

To-Do:
- [x] Get kicker firmware working again.
- [ ] ~~~Re-enable hardware buttons / check debug button logic.~~~ [On hold until hardware comes in.. may not have debug buttons?]
- [x] Add hw-test to simply program attiny for debug and training purposes.
- [x] Replace HackedKickerboard class with the real kickerboard class.
- [ ] ~~~Test code on kickerboard hardware.~~~ [On hold until hardware.]
    - [ ] ~~~Check debug buttons work as expected.~~~
    - [ ] ~~~Check that charging gets disabled when capacitor voltage is too high.~~~
    - [ ] ~~~Verify that SPI is no longer getting random interrupts from line noise.~~~

### DIFF
--- a/firmware/common2015/drivers/kicker-board/KickerBoard.cpp
+++ b/firmware/common2015/drivers/kicker-board/KickerBoard.cpp
@@ -1,6 +1,5 @@
 #include "KickerBoard.hpp"
 #include <tuple>
-#include "kicker_commands.h"
 
 using namespace std;
 
@@ -90,59 +89,51 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
     return true;
 }
 
-/* this function enforces the design choice that each cmd must have an arg
- * and return a value */
-uint8_t KickerBoard::send_to_kicker(const uint8_t cmd, const uint8_t arg,
-                                    bool verbose) {
+bool KickerBoard::send_to_kicker(uint8_t cmd, uint8_t arg, uint8_t& ret_val,
+                                 bool verbose) {
     chipSelect();
     // Returns state (charging, not charging), but we don't care about that
     uint8_t charge_resp = _spi->write(cmd);
     // Should return the command we just sent
     uint8_t command_resp = _spi->write(arg);
     // Should return final response to full cmd, arg pair
-    uint8_t return_resp = _spi->write(BLANK);
+    ret_val = _spi->write(BLANK);
     chipDeselect();
 
+    bool command_acked = command_resp == cmd;
     if (verbose) {
+        if (!command_acked) {
+            printf("Kicker failed to ack command! -- ");
+        }
+
         printf("Kicker: CHG:%02X, CMD:%02X, RET:%02X\r\n", charge_resp,
-               command_resp, return_resp);
+               command_resp, ret_val);
         fflush(stdout);
     }
 
-    return return_resp;
+    return command_acked;
 }
 
-uint8_t KickerBoard::kick(uint8_t time) {
-    return send_to_kicker(KICK_CMD, time);
+bool KickerBoard::send_to_kicker(uint8_t cmd, uint8_t arg, bool verbose) {
+    uint8_t val;
+    return send_to_kicker(cmd, arg, val, false);
 }
 
-uint8_t KickerBoard::chip(uint8_t time) {
-    return send_to_kicker(CHIP_CMD, time);
+bool KickerBoard::kick(uint8_t time) { return send_to_kicker(KICK_CMD, time); }
+
+bool KickerBoard::chip(uint8_t time) { return send_to_kicker(CHIP_CMD, time); }
+
+bool KickerBoard::read_voltage(uint8_t& voltage) {
+    return send_to_kicker(GET_VOLTAGE_CMD, BLANK, voltage);
 }
 
-uint8_t KickerBoard::read_voltage() {
-    return send_to_kicker(GET_VOLTAGE_CMD, BLANK);
-}
+bool KickerBoard::charge() { return send_to_kicker(SET_CHARGE_CMD, ON_ARG); }
 
-uint8_t KickerBoard::charge() { return send_to_kicker(SET_CHARGE_CMD, ON_ARG); }
-
-uint8_t KickerBoard::stop_charging() {
+bool KickerBoard::stop_charging() {
     return send_to_kicker(SET_CHARGE_CMD, OFF_ARG);
 }
 
-uint8_t KickerBoard::is_pingable() { return send_to_kicker(PING_CMD, BLANK); }
-
-uint8_t KickerBoard::is_kick_debug_pressed() {
-    return send_to_kicker(GET_BUTTON_STATE_CMD, DB_KICK_STATE);
-}
-
-uint8_t KickerBoard::is_chip_debug_pressed() {
-    return send_to_kicker(GET_BUTTON_STATE_CMD, DB_CHIP_STATE);
-}
-
-uint8_t KickerBoard::is_charge_debug_pressed() {
-    return send_to_kicker(GET_BUTTON_STATE_CMD, DB_CHARGE_STATE);
-}
+bool KickerBoard::is_pingable() { return send_to_kicker(PING_CMD, BLANK); }
 
 bool KickerBoard::is_charge_enabled() {
     uint8_t ret = 0;
--- a/firmware/common2015/drivers/kicker-board/KickerBoard.cpp
+++ b/firmware/common2015/drivers/kicker-board/KickerBoard.cpp
@@ -3,6 +3,9 @@
 
 using namespace std;
 
+
+std::shared_ptr<KickerBoard> KickerBoard::Instance;
+
 KickerBoard::KickerBoard(shared_ptr<SharedSPI> sharedSPI, PinName nCs,
                          PinName nReset, const string& progFilename)
     : AVR910(sharedSPI, nCs, nReset), _filename(progFilename) {}
@@ -81,6 +84,7 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
 }
 
 bool KickerBoard::send_to_kicker(uint8_t cmd, uint8_t arg, uint8_t* ret_val) {
+    LOG(INF2, "Sending: CMD:%02X, ARG:%02X", cmd, arg);
     chipSelect();
     // Returns state (charging, not charging), but we don't care about that
     uint8_t charge_resp = _spi->write(cmd);
--- a/firmware/common2015/drivers/kicker-board/KickerBoard.cpp
+++ b/firmware/common2015/drivers/kicker-board/KickerBoard.cpp
@@ -92,20 +92,24 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
 
 /* this function enforces the design choice that each cmd must have an arg
  * and return a value */
-uint8_t KickerBoard::send_to_kicker(const uint8_t cmd, const uint8_t arg) {
-    uint8_t val[2] = {0};
-    uint8_t ret = 0;
-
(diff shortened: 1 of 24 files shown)


jpfeltracco:
Okay, this is good to go.

It includes some fixes for our AVR programmer code (mainly related to using RTOS thread waits instead of normal MBED ones), a kickerboard class bug fix, tested attiny code as well as a hw-test that allows us to simply program an attiny using the AVR programming code directly. This test could be helpful for debugging, but it's also useful because we needed an easy way to program attiny chips for Robojackets electrical training (it's very simple to use the mbed as a programmer within our framework). Finally, the other hw-tests related to the kicker have been cleaned up.

It also adds back the basic kicker functionality in main.cpp that was in place before the hacked kickerboard was required.

This has been tested using a logic analyzer on a breadboarded MBED and functions correctly (as far as I can tell). Now the question is if we want to wait until we have the circuit boards ready and we test on the physical hardware before merging this in, or if we want to go ahead and do it on the sooner side. Since this has some changes to main.cpp my vote would be to merge it in sooner to avoid conflicts as we continue with more firmware work.

Thoughts? @jgkamat

P.S. I can't tell why checkstyle is failing, locally checkstyle is passing : /

jgkamat:
Don't worry about style for now, this is because the clang-format on your machine is probably newer than the CI machines. We'll fix that once this is reviewed.

I'll look over this eventually (but I don't know how much feedback I can give)

jpfeltracco:
Since firmware isn't your favorite do you want to pull in someone else for code review instead? It could take some burden off of you and help spread some knowledge around too.

jgkamat:
@JNeiger could you do the code review for this one? It would be awseome if you did =)

firmware/common2015/drivers/kicker-board/KickerBoard.cpp
```diff
@@ -89,4 +90,69 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
...
+
+uint8_t KickerBoard::chip(uint8_t time) {
+    return send_to_kicker(CHIP_CMD, time);
+}
+
+uint8_t KickerBoard::read_voltage() {
+    return send_to_kicker(GET_VOLTAGE_CMD, BLANK);
+}
+
+uint8_t KickerBoard::charge() { return send_to_kicker(SET_CHARGE_CMD, ON_ARG); }
```
JNeiger: I would split this into 3 lines instead of 1 so it matches the rest of the kicker board commands

firmware/common2015/drivers/kicker-board/KickerBoard.cpp
```diff
@@ -89,4 +90,69 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
...
 
+uint8_t KickerBoard::stop_charging() {
+    return send_to_kicker(SET_CHARGE_CMD, OFF_ARG);
+}
+
+uint8_t KickerBoard::is_pingable() { return send_to_kicker(PING_CMD, BLANK); }
```
JNeiger: Same as above, just split it into 3 lines.

firmware/common2015/drivers/kicker-board/KickerBoard.hpp
```diff
@@ -1,56 +1,117 @@
...
+     */
+    bool flash(bool onlyIfDifferent = true, bool verbose = false);
+
+    /**
+     * @brief Sends the KickerBoard a command to kick for the allotted time in
+     *     in milliseconds. This roughly corresponds to kick strength.
+     *
+     * @param time Millisecond kick time, can only range from 0 to 16 ms
+     */
+    uint8_t kick(uint8_t time);
```
JNeiger: Missing `@return` in comment

firmware/robot2015/hw-test/attiny-programmer-test.cpp
```diff
@@ -0,0 +1,55 @@
+// #include <rtos.h>
+
+// Doesn't help, tried above and below includes
+// #include <rtos.h>
+// #include "mbed.h"
+// #include "rtos.h"
+
+// #include "KickerBoard.hpp"
```
JNeiger: Left over includes probably should be removed

firmware/common2015/drivers/kicker-board/KickerBoard.hpp
```diff
@@ -1,56 +1,117 @@
...
+     * @param verbose If verbose, debug log messages are printed to stdout
+     * @return True if flashing was successful
+     */
+    bool flash(bool onlyIfDifferent = true, bool verbose = false);
+
+    /**
+     * @brief Sends the KickerBoard a command to kick for the allotted time in
+     *     in milliseconds. This roughly corresponds to kick strength.
+     *
+     * @param time Millisecond kick time, can only range from 0 to 16 ms
```
JNeiger: Is this comment updated? [Here](https://github.com/RoboJackets/robocup-firmware/pull/24/files#diff-f22dcee64f9b9230da09b03c98374d64R44) you are passing an argument that is greater than 16

firmware/common2015/drivers/kicker-board/KickerBoard.cpp
jpfeltracco: Unfortunately our checkstyle is too aggressive and forces one-lining this at the expense of readability IMO.

firmware/common2015/drivers/kicker-board/KickerBoard.hpp
jpfeltracco: Good catch, the protocol used to try to pack kick strength in the same byte containing the command rather than using separate bytes for each (now the limit is 0-255).

firmware/robot2015/src-kckr/main.c
```diff
@@ -1,13 +1,332 @@
...
+//         } else {
+//             execute_cmd(SET_CHARGE_CMD, ON_ARG);
+//         }
+//     }
+
+//     // Now our last state becomes the current state of the buttons
+//     kick_db_held_down_ = kick_db_pressed;
+//     chip_db_down_ = chip_db_pressed;
+//     charge_db_down_ = charge_db_pressed;
+// }
```
jpfeltracco: Possibly, we won't know until we have the hardware finished up I think.

firmware/robot2015/src-ctrl/main.cpp
```diff
@@ -135,12 +103,42 @@ int main() {
...
+    // Reprogramming each time (first arg of flash false) is actually
+    // faster than checking the full memory to see if we need to reflash.
+    bool kickerReady = kickerBoard.flash(false, true);
+
+    // flag fro kicking when the ball sense triggers
+    bool kickOnBreakBeam = false;
+    // Made up value right now, this is the amount of time in ms to
+    // allow the capacitor dump power into kicker. Will need to be
+    // adjusted once hardware is available.
+    uint8_t kickStrength = 5;
```
JNeiger: I think this is still on the old 0-16 system too

JNeiger:
Except for those couple things, everything else looks good!

firmware/robot2015/hw-test/attiny-programmer-test.cpp
```diff
@@ -0,0 +1,55 @@
...
+    pc.baud(57600);
+    shared_ptr<SharedSPI> sharedSPI =
+        make_shared<SharedSPI>(RJ_SPI_MOSI, RJ_SPI_MISO, RJ_SPI_SCK);
+    sharedSPI->format(8, 0);
+    FILE* fp = fopen(filename.c_str(), "r");
+    AVR910 programmer(sharedSPI, RJ_KICKER_nCS, RJ_KICKER_nRESET);
+
+    if (fp == NULL) {
+        pc.printf("Failed to open file.\r\n");
+        return 1;
```
JNeiger: This should probably be `-1` to follow normal conventions

jgkamat:
I'm personally for changing the style guides somehow to make these things 'valid'. Thoughts? 

https://github.com/RoboJackets/style-configs

We would have to coordinate with igvc and roboracing, but I don't think they would be opposed to nice things like this.

firmware/common2015/drivers/kicker-board/KickerBoard.cpp
```diff
@@ -89,4 +89,61 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
...
+    ret_val = _spi->write(BLANK);
+    chipDeselect();
+
+    bool command_acked = command_resp == cmd;
+    if (verbose) {
+        if (!command_acked) {
+            printf("Kicker failed to ack command! -- ");
+        }
+
+        printf("Kicker: CHG:%02X, CMD:%02X, RET:%02X\r\n", charge_resp,
```
JNeiger: These two `printf()` should be changed to `log()`. `printf()` doesn't play nicely with multiple threads and if two decide to try and print at the same time, their output becomes all jumbled.

jpfeltracco: Do you know what the behavior of LOG() will be when this is called from one of the hw-test files? The verbosity flags are almost exclusively used within the hw-tests and it would be nice if LOG() just printed the message over Serial like usual.

jpfeltracco:
@jgkamat I agree that it would be nice to accept both two line functions and one line functions.

The clang-format style documentation has an SFS_Inline option under AllowShortFunctionsOnASingleLine. You can read more [here](http://clang.llvm.org/docs/ClangFormatStyleOptions.html). Not sure if it's what we're looking for.

firmware/common2015/drivers/kicker-board/KickerBoard.cpp
JNeiger: Good point. I'm not completely sure how it will work within the hw-tests since I'm not too familiar with them. Both seem to be used interchangeably within the [radio-receiver-test](https://github.com/RoboJackets/robocup-firmware/blob/master/firmware/robot2015/hw-test/radio-receiver-test.cpp#L80). The only requirements for `LOG()` to work is for `isLogging` and `rjLogLevel` to be set. 

[`LOG()`](https://github.com/RoboJackets/robocup-firmware/blob/master/firmware/common2015/utils/logger/logger.cpp#L39) basically just surrounds the `printf()` with a mutex and adds a few "headers" such as time so it /should/ be the same as a normal `printf()`

If you want, the whole `if (verbose)` thing could be refactored to just `LOG(INF2, ...)` kinda like how it's done in the [mcp20317 class](https://github.com/RoboJackets/robocup-firmware/blob/b93c15c21918f78fd053ec0f146adade4f2cd878/firmware/common2015/drivers/mcp23017/mcp23017.cpp#L59). It looks to be the same type of data being printed out.

jpfeltracco:
My last two commits fixed up the issues @JNeiger brought up, fixed a few more style related issues, and changed over to preferring the logging system over printfs in threaded segments.

Before we merge this in, I would like to add in some debug commands for the kicker, I'll look into that tomorrow.

jpfeltracco:
Okay, full debug commands have been added so that we can type things like `kicker charge on` or `kicker kick` in the console.

There was another issue of getting junk SPI values back when used with the rest of our firmware on the shared SPI line (the hw-test was working fine). I think this was the result of the kicker getting interrupted by radio commands while not being chip-selected. We now totally disable the USI (module that handles SPI) when we aren't chip-selected, and clear the USI status register when we get chip-selected again. This appears to be working well now.

Now that this is all working, tested on a breadboard, and cleaned up, I'd say it's ready to merge in. We can make a separate issue for changing pin numbers around and verifying that this works with the actual kicker hardware once we have that circuit board in.

JNeiger:
Looks good to me! 👍

Issue #6, opened 2016-08-29
Firmware error logging to file


jgkamat:
Migrated from RoboJackets/robocup-software#316

JNeiger:
Just a quick update.
I have the basic write to file already done and completed. 

I am now trying to implement a system that will clear the file before it reaches the max file size and fills up the mbed internal flash. For some reason, any sort of file size implementation is returning seemingly random file sizes. I'm still working on it.

jgkamat:
That might be a little hard as I don't think there are any logging utils in the mbed libraries (if there are, you should probably use them). 

Have you tried this method of getting the size (this works in c: http://stackoverflow.com/questions/238603/how-can-i-get-a-files-size-in-c)

JNeiger:
I have tried that in the past, but I did not notice the fine print in the mbed file system page saying...

`fseek is not supported for files opened for writing ("w")`

If you try to do it while writing, it completely destroys the mbed flash memory and you have to reformat it.

Hopefully, opening it with `"a"` works with `fseek` and I can get around that restriction. I'll be trying that route later this week.

JNeiger:
With the incompatibilities between the file logger and the radio (Detailed in #26), is this still worth the added complexity to include?

The current implementation will only allow file logging while the radio is disabled.

guyfleeman:
This was replaced by flexible slots in the radio return packet, and is a another example of mbed incompatibility/immaturity. Closing.

PR #26, opened 2016-09-22
Firmware Logging to File

JNeiger:
Writes to a text file every log message, automatically resets the log when it hits ~500KB 
Issue: #6

### DIFF
--- a/firmware/common2015/utils/logger/logger.cpp
+++ b/firmware/common2015/utils/logger/logger.cpp
@@ -47,22 +47,22 @@ void log(uint8_t logLevel, const char* source, int line, const char* func,
         time_t sys_time = time(NULL);
         strftime(time_buf, 25, "%H:%M:%S", localtime(&sys_time));
 		
-		FILE *fp = fopen("/local/out.txt", "a");
-		int fileSize  = ftell(fp);
+        FILE *fp = fopen("/local/out.txt", "a");
+        int fileSize  = ftell(fp);
 
-		fseek(fp, 0L, SEEK_END);
+        fseek(fp, 0L, SEEK_END);
 		
-		if (fileSize > 500000)//Bytes
-		{
-			// Deletes contents of file
-			fclose(fp);
-			fp = fopen("/local/out.txt", "w");
-		}
-		else
-		{
-			// Returns to the correct location
-			fseek(fp, 0L, SEEK_SET);
-		}
+        if (fileSize > 500000)//Bytes
+        {
+            // Deletes contents of file
+            fclose(fp);
+            fp = fopen("/local/out.txt", "w");
+        }
+        else
+        {
+            // Returns to the correct location
+            fseek(fp, 0L, SEEK_SET);
+        }
 
 
         snprintf(newFormat, sizeof(newFormat),
@@ -73,11 +73,11 @@ void log(uint8_t logLevel, const char* source, int line, const char* func,
 
         fflush(stdout);
         vprintf(newFormat, args);
-		vfprintf(fp, newFormat, args);
+        vfprintf(fp, newFormat, args);
         fflush(stdout);
-		fflush(fp);
+        fflush(fp);
 
-		fclose(fp);
+        fclose(fp);
         va_end(args);
         log_mutex.unlock();
     }
--- a/firmware/common2015/utils/logger/logger.cpp
+++ b/firmware/common2015/utils/logger/logger.cpp
@@ -6,12 +6,11 @@
 #include <mbed.h>
 #include <rtos.h>
 #include <cstdio>
-#include <errno.h>
 
 const char* LOG_LEVEL_STRING[] = {FOREACH_LEVEL(GENERATE_STRING)};
 const char* LOG_FILE_NAMES[] = {"/local/LOG1.TXT", "/local/LOG2.TXT"};
 
-const int maxLogSize = 500000;  // bytes
+const int maxLogSize = 250000;  // 250 KB each (500 KB total)
 
 /* The initial logging level shows startup info along with any
  * warning messages [but hopefully there's none of those :) ].
@@ -19,11 +18,8 @@ const int maxLogSize = 500000;  // bytes
 bool isLogging;  // = RJ_LOGGING_EN;
 
 uint8_t rjLogLevel;
-uint8_t logNameInd = 255;
 
 Mutex log_mutex;
-
-void findSmallestLogFile();
 
 LogHelper::LogHelper(uint8_t logLevel, const char* source, int line,
                      const char* func) {
@@ -55,8 +51,11 @@ void log(uint8_t logLevel, const char* source, int line, const char* func,
         time_t sys_time = time(NULL);
         strftime(time_buf, 25, "%H:%M:%S", localtime(&sys_time));
 
-        // if first call to log, find the right file to edit
-        if (logNameInd > 1) findSmallestLogFile();
+        // If it's the first call to log(), find the right file to edit
+        static uint8_t logNameInd = 255;
+        if (logNameInd > 1) {
+            logNameInd = findSmallestLogFile();
+        }
 
         FILE* fp = fopen(LOG_FILE_NAMES[logNameInd], "a");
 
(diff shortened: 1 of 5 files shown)


jgkamat:
Looks pretty good! Does this delete the file every <x> bytes and start logging again? I'm not 100% sure right now.

JNeiger:
Yep!

jgkamat:
This /could/ cause us to have a mbed crash when we have minimal logs, but this is definitely better than nothing. Maybe we could make two files, and swap between them (so we always have one file of info)

@guyfleeman thoughs?

guyfleeman:
I like the idea of two files that switch or rollover that way we always keep a log. What are you concerned about causing a crash?

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -47,6 +47,19 @@ void log(uint8_t logLevel, const char* source, int line, const char* func,
         time_t sys_time = time(NULL);
         strftime(time_buf, 25, "%H:%M:%S", localtime(&sys_time));
 
+        FILE* fp = fopen("/local/log.txt", "a");
```
guyfleeman: This needs a NULL check

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -47,6 +47,19 @@ void log(uint8_t logLevel, const char* source, int line, const char* func,
...
+        fseek(fp, 0L, SEEK_END);
+
+        if (ftell(fp) > 500000)  // Bytes
+        {
+            // Deletes contents of file
+            fclose(fp);
+            fp = fopen("/local/log.txt", "w");
```
guyfleeman: same as above, add null checking

JNeiger:
Having it switch between files shouldn't be too hard to do.

JNeiger:
It writes to the first file until it reaches the max file size, then switches to file2. Once this one hits the max file size, it deletes the contents of the first file and begins to write to it again.

It is able to continue writing to the correct file through mbed restarts.

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -5,18 +5,26 @@
...
  * warning messages [but hopefully there's none of those :) ].
  */
 bool isLogging;  // = RJ_LOGGING_EN;
 
 uint8_t rjLogLevel;
+uint8_t logNameInd = 255;
 
 Mutex log_mutex;
 
+void findSmallestLogFile();
```
jgkamat: Could you put this in logger.hpp? I /think/ that function definitions are supposed to go in hpp files to speed up build times.

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -70,3 +105,19 @@ int logLvlChange(const std::string& s) {
 
     return n;
 }
+
+void findSmallestLogFile() {
```
jgkamat: Could you make this function return the logNameInd, and use it as a scoped variable above? It just seems kind of weird having it modify global variables.

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -47,6 +55,27 @@ void log(uint8_t logLevel, const char* source, int line, const char* func,
...
+        // if first call to log, find the right file to edit
+        if (logNameInd > 1) findSmallestLogFile();
+
+        FILE* fp = fopen(LOG_FILE_NAMES[logNameInd], "a");
+
+        if (fp) {
+            fseek(fp, 0L, SEEK_END);
+
+            if (ftell(fp) > maxLogSize) {
+                // Deletes contents of file
```
jgkamat: This comment implies that you are deleting the current fp... can you make it a bit more clear that you will overwrite the next opened file?

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -5,18 +5,26 @@
...
+
+const int maxLogSize = 500000;  // bytes
 
 /* The initial logging level shows startup info along with any
  * warning messages [but hopefully there's none of those :) ].
  */
 bool isLogging;  // = RJ_LOGGING_EN;
 
 uint8_t rjLogLevel;
+uint8_t logNameInd = 255;
```
jgkamat: If possible could you delete this and make logNameInd a static(?) variable inside the log function? See my comment in findSmallestLogFile for more info.

firmware/common2015/utils/logger/logger.cpp
```diff
@@ -5,18 +5,26 @@
 
 #include <mbed.h>
 #include <rtos.h>
+#include <cstdio>
+#include <errno.h>
 
 const char* LOG_LEVEL_STRING[] = {FOREACH_LEVEL(GENERATE_STRING)};
+const char* LOG_FILE_NAMES[] = {"/local/LOG1.TXT", "/local/LOG2.TXT"};
+
+const int maxLogSize = 500000;  // bytes
```
jgkamat: Could you put `500 KB` in this comment (so we don't have to look it up)

Also, with two log files this will take 1MB of space, is that acceptable? I'm not sure how many logs we generate in firmware...

JNeiger: The MBED only has 2 MB of space total, so I was trying to be a little conservative.
Overall, the startup sequence produces ~700 bytes worth of logs. And with my current setup, there are no periodic log calls.

JNeiger:
@jgkamat I added the bool disable flag. Can you double check it?

jgkamat:
LGTM! :+1: 

Be sure to double check it on an mbed though :smile:

jgkamat:
Let's merge this at tomorrow's meeting if no one objects...

jgkamat:
Ok, @JNeiger I think this is good. If you've tested this on an mbed and you think it works on all the edge cases, feel free to go ahead and merge this!

jjones646:
There's a good chance this will have to be disabled when receiving radio packets.

Because of the hardware layout, the mbed library [uses](https://github.com/ARMmbed/mbed-os/blob/3a16ca9855528fdb801b11f275ffdca7b5ed9f26/platform/semihost_api.h#L55) a [`bkpt`](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0552a/BABHCHGB.html) instruction for implementing the local file io. Basically, the processor is put in debug mode while the mbed's 2nd microcontroller retrieves & sends back the file from a flash chip. It's something that can really throw off the timing of everything, so just be weary of that.

jgkamat:
@JNeiger is it possible to store logs in ram and commit them to disk once we reach a certain amount? I'm not sure if we have enough memory for that though... 

There might not be an easy way to work around this (other than disabling it)

JNeiger:
@jjones646 If I'm understanding you right, any sort of local file io is going to have to be disabled whenever the radio is enabled, no matter the file io calling thread or frequency. 

Is timing problem because of a switch between debug mode and the normal mode? Or is there something in the underlying debug mode that will cause these timing issues?

jjones646:
@JNeiger The [`bkpt`](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0552a/BABHCHGB.html) instruction is literally a software breakpoint. So the processor doesn't continue until it's told to do so. There's really no way around it.

Run [pyocd-gdbserver](https://github.com/mbedmicro/pyOCD) with io code running at the start of main on an mbed (`make robot2015-gdb`). It'll stop once it hits [the](https://github.com/ARMmbed/mbed-os/blob/3a16ca9855528fdb801b11f275ffdca7b5ed9f26/platform/semihost_api.h#L55) instruction.

jgkamat:
@JNeiger since this is going to take a bit more time, can you merge master into this branch? I wanted to enable firmware build status on master, and merging master into this branch will prevent this branch's errors from showing up in status.

Thanks!

JNeiger:
@jgkamat Let me know if it's not, but it should be already updated. I merged master in last night and I can't find any updates since then.

jgkamat:
whoops, sorry! I didn't notice that you did that, my bad. You're good! =)

guyfleeman:
@JNeiger @jgkamat given the bkpt Jon mentioned above and recent additional feedback via radio, are we proceeding with the PR or will it live as debug branch or something?

JNeiger:
I would leave this as either a debug branch or get rid of it completely. TBH, I'm not sure how useful it is when you can't run it at the same time as the radio. There's not that much left that can be developed without having a radio

JNeiger:
I wrote this freshman year, imma call this dead lol

PR #28, opened 2016-11-21
Invalid encoder detection

jjones646:
### TODO:
- [ ] test

PR #29, opened 2017-01-15
Add support for the Decawave

jgkamat:
@petersonev 

Successor of #28 and includes its commits.

This needs to start getting reviewed since not using master is kinda hacky (and at that point we should probably just merge this in).


jgkamat:
The style errors on this one are pretty pervasive and result in a ton of changed lines, so it might be better to reveiw this as is, and apply style later.

Let me know when you think style should be fixed.

jgkamat:
Are the files in `firmware/common2015/drivers/decadriver` part of a library? It seems like a ton of code, and it might be better as part of a submodule.

petersonev:
All the files in decadriver are from decawave's website, not on any git repo. Also I've changed a decent bit in them and they're really only going to be used in our firmware so I don't see much of a reason to make it a submodule.

petersonev:
And ya, style makes a huge mess out of a lot of the decawave driver files so I don't know if we can adjust style in some way because for some parts I remember it trying to turn a couple lines into 20+ lines.

jgkamat:
We can turn of style selectively for some folders, so we could ignore all the decawave library folders. 

It might be possible to tune the style checker to try to do that, but we would have to give it different directives most likeley (making lines longer maybe). 

Worst case we can just ignore style for now, it will become green when it gets in master.

petersonev:
Okay, that'll be good. And all of the running of the robots for today has been on the latest version of this branch (with encoder detection merged in) and haven't come across any major unknown issues so should mostly be good to go.

petersonev:
Prevent style from editing `firmware/common2015/drivers/decawave/decadriver` and the rest should be good

petersonev:
Everything should be good to merge. Still some bugs to work out with radio stuff but better than how master is now.

jgkamat:
Going ahead and merging, just push to master if you fix bugs, unless they're big

PR #33, opened 2017-01-21
Misc firmware optimizations

petersonev:
Miscellaneous firmware optimizations, for memory in particular. Also adds option for release or debug flag when compiling to reduce program size when console and other stuff not needed. 

Needs more testing before merge but should mostly be ready to go. Known issue that there are sometimes radio RX issues on base station when compiling with debug flag.


jgkamat:
Just letting you know there's no way I can possibly review all of this (since I don't understand it, and it's massive), so feel free to merge when you feel it's ready.

PR #34, opened 2017-02-02
Add support for 16.10 in ubuntu-setup

jgkamat:
Port of robojackets/robocup-software#891 to firmware


jgkamat:
Shouldn't cause any issues. If you can't get firmware to build on 16.10 let me know.

PR #41, opened 2017-08-09
Update motor test

jpfeltracco:
Update motor test to work again after Jon and comp changes.

Should go in after comp2017


jpfeltracco:
We can wait, it isn't a big deal. And yup, `make robot2015-test-motor` will build it, there are a bunch of tests that likely don't compile right now listed in the makefile you could potentially start adding too.

PR #42, opened 2017-09-23
Restructure firmware repo

petersonev:
Restructured firmware repo to make more sense and removed old code from 2011. Not tested on robot yet but shouldn't have any problems as no firmware code was changed. Also added back osx support


joshhting:
Can you move FPGA.hpp and FPGA.cpp into the same file that would be great - Matt

PR #44, opened 2017-11-01
Wheel slip

### LINKED ISSUE #43: Identifying wheel slippage using just the encoders
See [this paper](http://people.idsia.ch/~foerster/2006/1/omnidrive_kiart_preprint.pdf) that we used for the `V_x, V_y and R_omega` to wheel velocities transformation.

image

where m is the wheel velocities vector, v translational and angular velocities, and D is the BotToWheel matrix in [this file](https://github.com/RoboJackets/robocup-firmware/blob/master/firmware/robot2015/src-ctrl/modules/control/RobotModel.hpp). D^+ is just the pseudoinverse.

If the last equation holds true, we should not have any wheel slippage. We can also simplify that last equation to 
image

JNeiger:
Closes #43 and adds in an extra wheel slippage mitigation tool.

### DIFF
--- a/firmware/robot2015/src-ctrl/main.cpp
+++ b/firmware/robot2015/src-ctrl/main.cpp
@@ -73,7 +73,7 @@ void Task_Simulate_RX_Packet(const void* args) {
 #endif
 
 void Task_Controller(const void* args);
-void Task_Controller_UpdateTarget(Eigen::Vector3f targetVel);
+void Task_Controller_UpdateTarget(Eigen::Vector3f targetVel, bool do_print);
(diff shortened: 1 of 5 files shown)


JNeiger:
I think this should be good to test on the robots again. That line where I find the kernel using LU factorization just returned a 0 vector every time.

Assuming that their observation that the Kernal of C_a was the row of the I-DD^+ matrix is correct, this should be the correct implementation. At the very least, this should do something now.

JNeiger:
Closed in favor of #45

PR #45, opened 2017-11-06
Fixed the spinning when accelerating sideways

JNeiger:


JNeiger:
There has been some segfaults when running, I'm not sure if it was happening before or if this caused it

PR #46, opened 2018-01-31
Upgrade MPU driver

jpfeltracco:
Upgrade MPU driver to [I2CDevLib's version](https://github.com/jrowberg/i2cdevlib/tree/master/Arduino/MPU6050).

This will be ready to go after a little more clean up.


jpfeltracco:
This should  be good to go now. It adds a calibration test and data printout test for the MPU. It doesn't change any of our main code right now, but it'll be used in the future.

PR #52, opened 2018-05-13
18.04 stable build

guyfleeman:
Made a package list for 18.04 that builds control,kicker,base targets on a clean 18.04 install.

16.04/18.04 packages diff:
```
27c27
< clang-3.6
---
> clang-3.9
30c30
< g++-4.9
---
> g++-5
33c33
< clang-format-3.6
---
> clang-format-3.9
```

MERGE AFTER COMP


guyfleeman:
Didn't see a tag for not ready/wait to merge. But this should maybe wait till after comp. It uses a separate package list so it shouldn't break any 16.04 install but the migration to g++-4.9 to g++-5.0 could be major in very subtle ways and I don't want that cropping up at comp.

jpfeltracco:
I'm fine merging this as long as we test on a bot first from 18.04. The GCC version bump would only be an issue for someone compiling firmware from an 18.04 system at comp right?

Also I'm not super worried about the GCC version change, I'm regularly compiling from GCC 7 and haven't noticed any issues.

@petersonev?

jgkamat:
< clang-3.6
    ---
    > clang-3.9

We can actually just say `clang` here, if we are willing to stop pinning the version. See [this diff](https://github.com/RoboJackets/robocup-software/pull/1107/files#diff-10f784dbc8ba0ce875144e63769422cbL32) for an example.

The only issue that I found with newer clang-format upgrades is that stylize no longer guesses the correct binary, which [I worked around by using `update-alternatives` to force the binary name 'clang-format' to exist](https://github.com/RoboJackets/robocup-software/blob/master/util/ubuntu-setup#L153). I don't think these changes should matter too much to us, outside of breaking style yet again...

guyfleeman:
Jeremy and I confirmed the gcc-arm-none-eabi is not generating valid binaries. Imo this isn't worth the risk pre comp

guyfleeman:
@jgkamat why did we start pinning versions in the first place?

jgkamat:
I'm not entirely sure, but I think we were using a PPA and the PPA was not providing an update for clang, but rather `clang-3.6`. Since `clang-3.6` was available on 14.04, we just always asked for that. Now that we don't use PPAs any more and we don't need a (unreasonable) minimum clang, we should be able to upgrade just fine.

We definitely shouldn't upgrade to 18.04 for now, but that will be a big issue next year. I'll make sure the backup images I bring to comp are 16.04 still :(

guyfleeman:
@jgkamat some people have fixed issues moving to arm-none-eabi-* v6.x.x by tweaking specs files. I have a diff between 4.9 and 6.3.1 but can't say I'm adept enough yet to read the file lol. Have you tweaked them before?

jgkamat:
I haven't even heard of those files so I'm probably not the best person to ask :P 

I can take a look later on and see if I come up with anything, but don't count on it...

On May 31, 2018 6:46:04 PM PDT, Will Stuckey <notifications@github.com> wrote:
>@jgkamat some people have fixed issues moving to arm-none-eabi-* v6.x.x
>by tweaking specs files. I have a diff between 4.9 and 6.3.1 but can't
>say I'm adept enough yet to read the file lol. Have you tweaked them
>before?

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

guyfleeman:
So this is a tad gross, but basically recent github issues seem to indicate GCC 4.9.x is the only supported compiler version for MBED OS 2. The latest commit will install the proper versions and ensure all bad versions are purged. Not all relevant packages are actually hard deps, so everything is purged and pinned.

Jay, I defs need some feedback on this one but I tested the script as much as possible (clean no existing toolchain, good toolchain, and bad toolchain). It fixed the issue in all cases.

Control board firmware now boots, lights flash, init checks complete, and radio transmits. Probably needs a full run just to be extra safe but ultimately the version didn't move, we just messed with deps.

guyfleeman:
@jgkamat did you get any more info from your Ubuntu/Debian friends?

guyfleeman:
I'd appreciate some nitpicks from an experienced scripter :)

guyfleeman:
@jgkamat can you trigger a rebuild on this? Error looks unrelated to what I added?

jgkamat:
> @jgkamat can you trigger a rebuild on this? Error looks unrelated to what I added?

Yup, here was the error for reference:

Collecting websocket-client (from pyocd->-r /home/developer/project/util/requirements2.txt (line 10))
Exception:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 209, in main
    status = self.run(options, args)
  File "/usr/lib/python2.7/dist-packages/pip/commands/install.py", line 317, in run
    requirement_set.prepare_files(finder)
  File "/usr/lib/python2.7/dist-packages/pip/req/req_set.py", line 360, in prepare_files
    ignore_dependencies=self.ignore_dependencies))
  File "/usr/lib/python2.7/dist-packages/pip/req/req_set.py", line 577, in _prepare_file
    session=self.session, hashes=hashes)
  File "/usr/lib/python2.7/dist-packages/pip/download.py", line 810, in unpack_url
    hashes=hashes
  File "/usr/lib/python2.7/dist-packages/pip/download.py", line 649, in unpack_http_url
    hashes)
  File "/usr/lib/python2.7/dist-packages/pip/download.py", line 842, in _download_http_url
    stream=True,
  File "/usr/share/python-wheels/requests-2.9.1-py2.py3-none-any.whl/requests/sessions.py", line 480, in get
    return self.request('GET', url, **kwargs)
  File "/usr/lib/python2.7/dist-packages/pip/download.py", line 378, in request
    return super(PipSession, self).request(method, url, *args, **kwargs)
  File "/usr/share/python-wheels/requests-2.9.1-py2.py3-none-any.whl/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/share/python-wheels/requests-2.9.1-py2.py3-none-any.whl/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/share/python-wheels/CacheControl-0.11.5-py2.py3-none-any.whl/cachecontrol/adapter.py", line 46, in send
    resp = super(CacheControlAdapter, self).send(request, **kw)
  File "/usr/share/python-wheels/requests-2.9.1-py2.py3-none-any.whl/requests/adapters.py", line 376, in send
    timeout=timeout
  File "/usr/share/python-wheels/urllib3-1.13.1-py2.py3-none-any.whl/urllib3/connectionpool.py", line 610, in urlopen
    _stacktrace=sys.exc_info()[2])
  File "/usr/share/python-wheels/urllib3-1.13.1-py2.py3-none-any.whl/urllib3/util/retry.py", line 228, in increment
    total -= 1
TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'

Looks like an error in installing pyocd. It seems to install cleanly on my
machine, hopefully it was just a temporary issue (maybe a bad
'websocket-client' update). Otherwise we may need to
look at that more (I can do that on fri or later).

Will Stuckey writes:

util/ubuntu-setup
```diff
@@ -73,6 +76,34 @@ if [ $UID -ne 0 ]; then
...
+		echo "cannot continue on Bionic without permission to add to lists"
+		echo "-- exit"
+		exit 1
+        	;;
+	esac
+fi
+
+if [ "ubuntu-18.04" = "$SYSTEM" ]; then
+	SOURCE_LIST=/etc/apt/sources.list.d/gcc-arm-none-eabi_xenial.list
+	if [ ! -f $SOURCE_LIST ]; then
```
jgkamat: Usually when expanding variables we would quote them, to avoid parsing spaces as multiple arguments, like this:

if [ ! -f "$SOURCES_LIST" ]; then

however, in this case, it dosen't matter since the variable is hard coded. I would still prefer this var to be quoted, just in case the path gets a space at some point (and just because it's good practice).

util/ubuntu-setup
```diff
@@ -92,11 +123,40 @@ if $YES; then
...
+		echo "-- found existing ARM compiler install, verifying"
+	fi
+
+	# compiler is installed, check if the version is correct
+	if [ "$ARM_COMPILER" = true ]; then
+		if [[ $(apt-cache policy gcc-arm-none-eabi) = *"Installed: 15:4.9.3"* ]]; then
+			echo "-- detected good arm compiler version"
+		else
+			echo "-- detected bad arm compiler verison, purging"
+			apt-get purge $TOOLCHAIN_PACKAGES $ARGS
```
jgkamat: Just because I made the above comment, this variable should *not* be quoted since we're counting on the spaces being evaluated. So this is good :)

util/ubuntu-setup
```diff
@@ -92,11 +123,40 @@ if $YES; then
     ARGS="-y"
 fi
 
-PACKAGES="$(sed 's/#.*//;/^$/d' $BASE/util/ubuntu-packages.txt)"
+TOOLCHAIN_PACKAGES="libnewlib-dev libnewlib-arm-none-eabi gcc-arm-none-eabi libstdc++-arm-none-eabi-newlib"
+if [ "ubuntu-18.04" = "$SYSTEM" ]; then
+	# check for missing compiler
+	ARM_COMPILER=true
+	if [[ $(apt-cache policy gcc-arm-none-eabi) = *"Installed: (none)"* ]]; then
```
jgkamat: Checking if the package is installed is probably better done by inspecting the exit code of

dpkg -s gcc-arm-none-eabi 

It seems to work for me at least :P

util/ubuntu-setup
```diff
@@ -73,6 +76,34 @@ if [ $UID -ne 0 ]; then
...
+        	;;
+	esac
+fi
+
+if [ "ubuntu-18.04" = "$SYSTEM" ]; then
+	SOURCE_LIST=/etc/apt/sources.list.d/gcc-arm-none-eabi_xenial.list
+	if [ ! -f $SOURCE_LIST ]; then
+		echo "-- adding xenial universe to support arm compiler"
+		touch $SOURCE_LIST
+		echo "# robocup-firmware added xenial universe for gcc-arm-none-eabi 5.4.1" > $SOURCE_LIST
```
jgkamat: This can be much simpler with a [heredoc](http://www.tldp.org/LDP/abs/html/here-docs.html)

Also, you probably meant to do `>>` a line below, but it won't matter if you switch to heredocs :P

guyfleeman:
ssl error on build this time, also appears unrelated to changes

jgkamat:
> ssl error on build this time, also appears unrelated to changes

It's odd that this happens whenever we aren't triggering a rebuild (I've seen
it happen on a couple other builds on software as well) :P

Maybe there's some infra error over at pypi or something.

It shouldn't affect us as long as we are using cached builds (unfortunately
this PR is breaking the cache a lot, so it will see it more than most).

PR #58, opened 2018-06-06
Allow low millisecond kicks

jpfeltracco:


src/kicker/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ set(CMAKE_CXX_COMPILER  ${AVR_CXX_COMPILER})
 # change above to match ATTINY84A =====================================================================
 # to keep timing for neopixel
 # The -O2 turns on optimizations, which are required to use the delay_*() functions
-list(APPEND CMAKE_C_FLAGS "-std=c11 -O2 -mmcu=attiny167 -DF_CPU=1000000")
```
ashaw596: Just wondering. Why did this change?

guyfleeman: As far as we can tell it was never correct. It doesn't actually affect the clock or core timer scalers, just some delay/wait functions.

guyfleeman:
Admittedly the timing is a little jank, but this is largely due to the limit prescale choices on the timer of 1,8,32... So this is the most practical setting.

guyfleeman:
This was confirmed functioning on hardware.

src/kicker/main.c
```diff
@@ -52,19 +55,19 @@ unsigned time = 0;
...
-    pre_kick_cooldown_ = 5;
+    pre_kick_cooldown_ = 5 * TIMER_TO_MS;
     // minimum of 6 ms, we were breaking kickers with low duty cycles
-    // maximum of 6 + 7 == 13 ms
-    // millis_left_ = (int) ((strength / 255.0) * 6.0) + 7;
-    millis_left_ = 13;  // always full kick speed
-    post_kick_cooldown_ = 5;
-    kick_wait = 2000;
+    float time_cnt_flt = ((strength / 255.0f) * 9.0f + 0.8f) * TIMER_TO_MS;
+    timer_cnts_left_ = (int) (time_cnt_flt + 0.5f);
```
ashaw596: Why are we adding this here?

src/kicker/main.c
```diff
@@ -10,15 +10,18 @@
 
 #define NO_COMMAND 0
 
-#define TIMING_CONSTANT 125
+#define TIMING_CONSTANT 10
 #define VOLTAGE_READ_DELAY_MS 40
 
+// 1 ms / 80 us = 12.5
+#define TIMER_TO_MS 12.5f
```
ashaw596: If we're using an 8Mhz with 8 scale and a 10 timing constant. shouldn't this be 8e6/8/10 = 1e5 timer period = 
1e-5 timer period. 1 ms = 100 timer.  Also shouldn't this be called TIMER_TO_MS?

Not sure, i think thats rightish. If it is that timing constant seems too low.

src/kicker/main.c
```diff
@@ -139,7 +142,7 @@ void main() {
             kick_on_breakbeam_ = false;
         }
 
-        _delay_us(100);  // 0.1 ms
+        _delay_us(10);  // 0.1 ms
```
ashaw596: This delays 100 ms

src/kicker/main.c
ashaw596: Oh whoops. could you add a comment for rounding?

src/kicker/main.c
```diff
@@ -52,19 +55,19 @@ unsigned time = 0;
...
-    return pre_kick_cooldown_ || millis_left_ || post_kick_cooldown_ ||
+    return pre_kick_cooldown_ || timer_cnts_left_ || post_kick_cooldown_ ||
            kick_wait;
 }
 
 void kick(uint8_t strength) {
     if (is_kicking()) return;
-    pre_kick_cooldown_ = 5;
+    pre_kick_cooldown_ = 5 * TIMER_TO_MS;
     // minimum of 6 ms, we were breaking kickers with low duty cycles
```
ashaw596: x

src/kicker/main.c
ashaw596: i mean 0.01 ms. whoops

src/kicker/main.c
```diff
@@ -184,7 +187,8 @@ void init() {
 
     // OCR0A is max val of timer before reset
     // we need 1000 clocks at 1 Mhz to get 1 millisecond
-    // if we prescale by 8, then we need 125 on timer to get 1 ms exactly
+    // ~80 us - we think - Will
+    // don't worry about it - Jeremy
     OCR0A = TIMING_CONSTANT;  // reset every millisecond
```
ashaw596: The comment on this line is wrong right?

jpfeltracco:
@ashaw596 Yeah, this was jank put together last night. I'll clean it up before a merge.

guyfleeman:
@jpfeltracco are you busy? I was just about to clean and fix style.

src/kicker/main.c
guyfleeman: yea, it's 80us as indicated by ln 190

src/kicker/main.c
guyfleeman: yup

jpfeltracco:
I'm heading out for a bit at 10:40, so I'll just commit and push the little stuff I changed and let you do the rest if you'd like.

It definitely still seems like we're missing another div by 8 somewhere though. We may want to explicitly set the clock prescale to 0 to make sure it doesn't have some weird default.

src/kicker/main.c
guyfleeman: So the old math also didn't make sense. I think there's an implicit pre-timer prescale of 8 (8Mhz->1MHz) for TIMER0. Then for the counter, we set the timer prescale to 8 (125KHz). We used to count to 125, giving a callback of 1KHz, or every ms. By dropping the count to 10, we get 80 us callbacks. 80us * 12.5 = 1ms. basically by doing this we've just allowed fractions of a ms resolution rather than ms resolution. 

As per the above formula, I think the name is okay. It's ms = timer_val * TIMER_TO_MS

guyfleeman:
I'll look for global prescale register and set it if it exists @jpfeltracco

src/kicker/main.c
ashaw596: Well. In that case, the way you're using it doesn't make sense.

float time_cnt_flt = ((strength / 255.0f) * 9.0f + 0.8f) * TIMER_TO_MS;
You're doing
timer = ms * TIMER_TO_MS

I assume. I think the float is in MS?
=((( I miss the c++ chrono library. static checking for time conversion plzzz

 I didn't see anywhere where the TIMER has an implicit prescale already. I think.
Just going off random blog posts tho. http://maxembedded.com/2011/06/avr-timers-timer0/

What did we setup the attiny clock to be? It has a couple settings from 1 MHz to 8 right? maybe. Embedded so complicatedddd

guyfleeman:
@jpfeltracco added comments and removed magic numbers

guyfleeman:
Kicker fault flickers on soccer after kick then goes away.

jpfeltracco:
Was that introduced by our recent changes?

jpfeltracco:
I think I found where that missing /8 came from: 

The CKDIV8 Fuse determines the initial value 
of the CLKPS bits. If CKDIV8 is unprogrammed,
the CLKPS bits will be reset to “0000”. If CKDIV8 is programmed, CLKPS bits are reset to
“0011”, giving a division factor of eight at star
t up. This feature should be used if the selected
clock source has a higher frequency than the maximum frequency of the device at the present
operating conditions. Note that any value can be written to the CLKPS bits regardless of the
CKDIV8 Fuse setting. The Application software must ensure that a sufficient division factor is
chosen if the selected clock source has a higher frequency than the maximum frequency of the
device at the present operating conditions. The device is shipped with the CKDIV8 Fuse
programmed.

Pg. 38 of the datasheet.

src/kicker/main.c
```diff
@@ -121,13 +126,15 @@ void main() {
         // get a voltage reading by weighing in a new reading, same concept as
         // TCP RTT estimates (exponentially weighted sum)
 
-        if (time % 400 == 0) {
+        // 10 us * 4000 = 40 ms, we don't need to run the ADC much faster
+        if (time % 4000 == 0) {
```
ashaw596: Not 100% sure how time works, but is this only giving a 10 us window for this to run every 40 ms? and we have a 10 us delay in this loop. Maybe store the last time it ran?

src/kicker/main.c
```diff
@@ -317,21 +332,22 @@ ISR(PCINT0_vect) {
  * (When CS00 is one, and CS01 is 0, no prescale. We don't use this)
  */
 ISR(TIMER0_COMPA_vect) {
```
ashaw596: This seems like alot of code to call every 80 processor cycles. Do we need that granular timing.

src/kicker/main.c
```diff
@@ -317,21 +332,22 @@ ISR(PCINT0_vect) {
...
+    sei();
```
ashaw596: Uhhh. This enables interrupts. I think

src/kicker/main.c
```diff
@@ -366,6 +382,7 @@ ISR(TIMER0_COMPA_vect) {
         // stop prescaled timer
         TCCR0B &= ~_BV(CS01);
     }
+    cli();
```
ashaw596: This disables interupts. Has this code been workign????

src/kicker/main.c
```diff
@@ -46,27 +63,43 @@ volatile uint8_t kick_on_breakbeam_strength_ = 0;
 
 unsigned ball_sense_change_count_ = 0;
 
-unsigned time = 0;
+uint32_t time = 0;
```
ashaw596: Wait Seriously. How is this updated? I can't find it anywhere.

src/kicker/main.c
jpfeltracco: That was a temporary (and broken) commit that I was working on, I'll let you know when this is ready for a merge.

src/kicker/main.c
jpfeltracco: Yup, that's how you enable nested interrupts, we were trying to allow spi communication to interrupt the timer to prevent a separate issue

src/kicker/main.c
jpfeltracco: Probably not, but all this chip has to do during the kick cycle is time the kicking. You don't need ballsense or voltage monitoring during that time.

src/kicker/main.c
jpfeltracco: The intention was just to sporadically call the ADC read.

src/kicker/main.c
jpfeltracco: That is a legitimate bug actually, time++ disappeared somewhere, so I guess we're constantly calling the ADC read which could throw off some timing in the main loop. Thanks!

src/kicker/main.c
```diff
@@ -8,17 +8,28 @@
...
+#define POST_KICK_SAFETY_MARGIN_MS 5
+
 #define NO_COMMAND 0
 
-#define TIMING_CONSTANT 125
+#define TIMING_CONSTANT 10
 #define VOLTAGE_READ_DELAY_MS 40
 
+// 1 ms / 80 us = 12.5
+#define MS_TO_TIMER 12.5f
```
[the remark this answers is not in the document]
jpfeltracco: Sure, I'll throw the math in a #define.

Keep in mind that you are reviewing old code here. The processor was only running in a 1 Mhz mode, now we've bumped it to 8 Mhz and have a MS_TO_TIMER value of 100.

We also have it hooked up to an o-scope to verify.

PR #62, opened 2018-06-14
IMU Integration + Gyro Angle Hold

jpfeltracco:
This PR adds our first IMU integration in firmware. We use the rotational velocity measurement from the gyro to stabilize our bot's rotation when soccer is commanding little or no rotational velocity itself.

Nothing has changed about the command packet structure, and this only activates when we are trying to strafe or otherwise move without intentional rotation. Further PRs will likely move all rotation control to the bot itself, and send world velocities rather than body relative so the bot's faster inner control loop can handle divergences from that much more quickly.

**Operation Changes**:
The IMU requires calibration to function without drift. A calibration test program is already in our repo and can be compiled + flashed using `make robot-test-imu-calibration-prog`. This will write a one line file with 6 offsets to `offsets.txt` on the mbed's file system, which will in turn be read on start up the the control board firmware.

**Related Issues**:
- @guyfleeman refactored the way I2C is handled to avoid thread safety issues in this branch
- For some reason, when this branch is compiled on 16.04 or other, newer versions of the arm toolcahin (like on Arch), the binaries produced segfault regularly.
- File system IO was causing segfaults in the PidMotionController's constructor or ControllerTaskThread's initialization. Therefore it was moved into main and passed all the way through.

### DIFF
--- a/src/control/modules/motion-control/PidMotionController.hpp
+++ b/src/control/modules/motion-control/PidMotionController.hpp
@@ -123,7 +123,7 @@ public:
         float target_w = _targetVel[2];
         target_w = rotation_pid.run(rotation); // rotation pid to do an angular hold
         // correct target velocity to include rotation hold
-        _targetVel[2] += target_w;
+        _targetVel[2] -= target_w;
 
         // conversion to commanded wheel velocities
         Eigen::Vector4d targetWheelVels = RobotModel::get().BotToWheel * _targetVel.cast<double>();
--- a/src/control/modules/motion-control/PidMotionController.hpp
+++ b/src/control/modules/motion-control/PidMotionController.hpp
@@ -15,7 +15,7 @@ class PidMotionController {
 public:
     PidMotionController() : imu(shared_i2c, MPU6050_DEFAULT_ADDRESS),
         ax_offset(0), ay_offset(0), az_offset(0), gx_offset(0), gy_offset(0), gz_offset(0),
-        ax(0), ay(0), az(0), gx(0), gy(0), gz(0), rotation(0), angular_vel(0)
+        ax(0), ay(0), az(0), gx(0), gy(0), gz(0), rotation(0), angular_vel(0), target_rotation(0)
     { 
         setPidValues(1.0, 0, 0, 50, 0);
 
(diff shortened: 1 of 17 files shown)


ashaw596:
What is SharedI2C for? It doesn't seem to do anything.

ashaw596:
Whoops nvm.

src/control/modules/motion-control/PidMotionController.hpp
```diff
@@ -56,14 +102,85 @@ class PidMotionController {
...
+        bool soccer_stop = (std::abs(target_vel_act[0]) < epsilon) &&
+                           (std::abs(target_vel_act[1]) < epsilon);
+        if (!soccer_stop && std::abs(target_vel_act[2]) < epsilon) {
+            if (!angle_hold) {
+                target_rotation = rotation;
+                angle_hold = true;
+            }
+
+            // get the smallest difference between two angles
+            float rot_error = std::atan2(std::sin(target_rotation - rotation),
```
ashaw596: It might not matter. But the atan is probably a really computationally expensive way to do this especially since we don't have native floating point support. Probably. I think.

Maybe more ridiculous, but something like this might be faster.
double difference = secondAngle - firstAngle;
        while (difference < -180) difference += 360;
        while (difference > 180) difference -= 360;

src/control/modules/motion-control/PidMotionController.hpp
```diff
@@ -56,14 +102,85 @@ class PidMotionController {
...
         Eigen::Vector4d wheelVels;
         wheelVels << encoderDeltas[0], encoderDeltas[1], encoderDeltas[2],
             encoderDeltas[3];
         wheelVels *= 2.0 * M_PI / ENC_TICKS_PER_TURN / dt;
 
+        auto bot_vel = RobotModel::get().WheelToBot * wheelVels;
+
+        // we have two redundent sensor measurements for rotation
+        // 32.8 comes from data sheet, units are LSB / (deg/s)
+        float ang_vel_gyro = (gz / 32.8f) * M_PI / 180.0f;
```
JNeiger: The typical scale factors are usually ok for the Gyro/Accelerometers, but if it becomes a problem, we can do some calibration to estimate those too. It's a pretty straightforward process

PR #63, opened 2018-06-16
Will/safe kicker charge (fixed #61)

### LINKED ISSUE #61: ATtiny Should Detect Kicker Hardware Failure
Right now once the attiny is flashed, if will continue to assert the charge line even if the voltage remains at zero. Charging is prohibited during kicks, and discharge occurs with no power. If we are attempting to charge and voltage doesn't increase we have a short on the high voltage DC side to ground, or a component(s) in the boost converter has failed. Many boost converter failures are destructive and will go undetected in a match. The ATtiny should deassert the charge pin and enter a changing error state until reboot. 

I didn't see any checks in the Kicker board.cpp driver on the mbed side but regardless this should happen on the ATtiny side for obvious safety reasons.

guyfleeman:


guyfleeman:
This is addressed partially in hardware kicker-v3.0. Not going to backport.

PR #68, opened 2019-01-20
New Radio Driver

CollinAvidano:
PR for the new radio driver. Current state is we can send and receive with issues on receiving.


lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+#define CMD_SET_SECURITY "C3="
+#define CMD_SET_DHCP "C4="
+#define CMD_GET_CONNECTION_INFO "C?"
+
+#define CMD_SET_TRANSPORT_PROTOCOL "P1="
+#define CMD_SET_HOST_IP "P3="
+#define CMD_SET_PORT "P4="
+#define CMD_START_CLIENT "P6="
+
+#define CMD_SEND_DATA "S3="
```
JNeiger: Does every command have a "XX=" format?

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+Ism43340::Ism43340(SpiPtrT sharedSPI, PinName nCs, PinName intPin,
```
JNeiger: This camel case device name style kills me

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+    while (dataready.read() != 1) {
+    }
+
+    uint8_t lastc = 0;
+    radiospi.chipselect();
+    for (int i = 0; i < command->size(); i++) {
+        uint8_t c = command[i];
+        if (lastc != 0) {
+            uint16_t packet = (c << 8) | lastc;
+            radiospi.m_spi->write(packet);
```
JNeiger: Is there a multi write byte? There may be some underlying differences in performance

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+int Ism43340::writetospi(std::string* command) {
+    while (dataready.read() != 1) {
+    }
+
+    uint8_t lastc = 0;
+    radiospi.chipselect();
+    for (int i = 0; i < command->size(); i++) {
+        uint8_t c = command[i];
+        if (lastc != 0) {
```
JNeiger: Is there ever a chance you can have '\0' in the data being sent? This would fail in that case

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+int Ism43340::readfromspi(uint8* readbuffer, uint32* readlength) {
+    while (dataready.read() != 1) {
+    }
+
+    packetcount = 0;
+    radiospi.chipselect();
+    while (dataready.read() != 0) {
+        uint16_t data = radiospi.m_spi->write(0x0a0a);
```
JNeiger: I would make this a define

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+    // you could probably have read spi return an int as a status code for both cant decide
+}
+
+// This command will probably not be used but is here as an example command
+// Set Human Readable
+inline setHumanReadable() {
+    std::string compositeString(CMD_SET_HUMAN_READABLE);
+
+    // formats command adding /r and /n when needed
+    formatCommand(&compositeString);
```
JNeiger: Are there any reasons that a command wouldn't have this? It should almost be in a lower function

lib/drivers/ism43340/ism43340.cpp
JNeiger: You can change the lastc to be signed and slightly larger, then make the default -1

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+inline void Ism43340::formatCommand(std::string* command) {
+    if (command->size() % 2 == 0) {
+        command->append("\r\n");
+    } else {
+        command->append("\r");
+    }
+}
+
+int Ism43340::writetospi(std::string* command) {
```
JNeiger: Use & instead of *

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+int32_t Ism43340::sendPacket(const rtp::Packet* pkt) {
+    // Return failutre if not initialized
+    if (!m_isInit) return COMM_FAILURE;
+
+
+    // dont know how much of this needs to be kept. Soccer may actually want the split between body and head of the sent packet. I dont know
+    BufferT txBuffer;
+
+    const auto bufferSize = 9 + pkt->size() + 2;
```
JNeiger: I like making these defines or const uints

lib/drivers/ism43340/ism43340.cpp
```diff
@@ -0,0 +1,270 @@
...
+    txBuffer.reserve(bufferSize);
+
+    // MAC layer header for Decawave
+    txBuffer.insert(txBuffer.end(),
+                    {0x41, 0x88, 0x00, 0xCA, 0xDE, pkt->header.address, 0x00,
```
JNeiger: I like defines for these too

lib/drivers/ism43340/ism43340.cpp
guyfleeman: I think this is an endianness isuees. The radio is the opposite endianness of the of mbed. I believe there is a write_8 and write_16 (hidden) but the endianness is wrong and the hardware module doesn't support switching.

lib/drivers/ism43340/ism43340.cpp
guyfleeman: Agreed. Is this left over cruft from the DWM packet header?

lib/drivers/ism43340/ism43340.cpp
guyfleeman: This is a DWM thing, I think it'll go away. Note the comment says MAC layer for DWM

lib/drivers/ism43340/ism43340.cpp
CollinAvidano: Now that my life is starting to actually come back together somewhat I just wanted to comment on some of these. Multibyte write through the mbeds spi api only supports 8 bit transfers and as such wont do any endianness conversion.

lib/drivers/ism43340/ism43340.cpp
guyfleeman: Yeah null bytes can def be a thing. We will need hard code max buffer length, and perhaps encode total length in the packet header and stop short based on the transmission length. Although, I don't anticipate the forward packets changing length, on the return when we wanna send telem

PR #70, opened 2019-04-28
Flash and test FPGA and Kicker

guyfleeman:


guyfleeman:
should this go to master?

PR #71, opened 2019-04-28
Math for mTrain motion control

kylestach:
None of this has been tested, and it's incomplete and it's still definitely a work in progress(this just does the math, we still need to write an interface between this and the motors/encoders/etc). I'm putting it out for review sooner rather than later so other motion control people can take a look, mostly to keep us all on the same page.

That said, it's mostly a port of the python code I prototyped at https://github.com/RoboJackets/robocup-analysis/tree/torque_control_simulation/model-python (other than the estimator, which should be pretty straightforward).

### DIFF
--- a/robot/control/modules/motion-control/MotionControl.hpp
+++ b/robot/control/modules/motion-control/MotionControl.hpp
@@ -22,10 +22,10 @@ class MotionControl {
 public:
     // Called at 1000hz to update the filter based on encoder values
     // and calculate torque goals.
-    Vector<4> update_control(Vector<5> odometry, time_point now);
+    MotorVoltage update_control(Odometry odometry, time_point now);
 
     // Called at ~60hz to update with new camera input.
-    void update_camera(Vector<3> camera, time_point camera_time, time_point now);
+    void update_camera(Camera camera, time_point camera_time, time_point now);
 
     void set_waypoint(Waypoint waypoint);
 
--- a/robot/control/modules/motion-control/MotionControl.hpp
+++ b/robot/control/modules/motion-control/MotionControl.hpp
@@ -1,0 +1,41 @@
+#pragma once
+
+#include <chrono>
+
+#include "RobotModel.hpp"
+#include "RobotEstimator.hpp"
+#include "RobotController.hpp"
+
+namespace motion {
+
+using clock = std::chrono::steady_clock;
+using time_point = std::chrono::time_point<clock>;
+
+struct Waypoint {
+    RobotPose pose;
+    RobotTwist velocity;
+    RobotTwist acceleration;
+    time_point time;
+};
+
+class MotionControl {
+public:
+    // Called at 1000hz to update the filter based on encoder values
+    // and calculate torque goals.
+    Vector<4> update_control(Vector<5> odometry, time_point now);
+
+    // Called at ~60hz to update with new camera input.
+    void update_camera(Vector<3> camera, time_point camera_time, time_point now);
+
+    void set_waypoint(Waypoint waypoint);
+
+private:
+    RobotEstimator estimator{RobotPose::Zero(), RobotTwist::Zero()};
+    RobotController controller;
+
+    MotorVoltage last_u = MotorVoltage::Zero();
+
+    Waypoint waypoint;
+};
+
+};
--- a/robot/control/modules/motion-control/RobotEstimator.cpp
+++ b/robot/control/modules/motion-control/RobotEstimator.cpp
@@ -28,20 +28,20 @@ void RobotEstimator::predict(Vector<4> u) {
     x.block<3, 1>(3, 0) += acceleration * kLoopTime;
 }
 
-void RobotEstimator::update_odometry(Vector<5> odometry) {
+void RobotEstimator::update_odometry(Odometry odometry) {
     // Rotate into the body frame to do the Kalman update,
     // then rotate back to apply changes.
     Matrix<3> rotation_matrix = make_rotation_matrix(x(2));
 
     RobotTwist body_velocity = rotation_matrix.transpose() * x.block<3, 1>(3, 0);
-    Vector<5> expected = robot_model.read_odom(body_velocity);
+    Odometry expected = robot_model.read_odom(body_velocity);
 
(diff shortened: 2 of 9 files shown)


robot/control/modules/motion-control/RobotModel.cpp
```diff
@@ -0,0 +1,134 @@
...
+    Vector<4> wheel_torque = motor_torque / kGearRatio;
+    Vector<3> effort = G.transpose() * wheel_torque;
+
+    Matrix<3, 3> coriolis = (Matrix<3, 3>() <<
+        0, -v(2), 0,
+        v(2), 0, 0,
+        0, 0, 0
+        ).finished();
+
+    Matrix<3, 3> V = wheel_friction_mat + coriolis * M;
```
JNeiger: Is this actually the coriolis force cross product?

robot/control/modules/motion-control/MotionControl.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <chrono>
+
+#include "RobotModel.hpp"
+#include "RobotEstimator.hpp"
+#include "RobotController.hpp"
+
+namespace motion {
+
+using clock = std::chrono::steady_clock;
+using time_point = std::chrono::time_point<clock>;
```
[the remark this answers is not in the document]
kylestach: Normally I'd agree with you, but it's within the `motion` namespace so it doesn't pollute the global namespace which is a case I think it's reasonable to do it for clarity. The other, more specific reason I used it here is to abstract the the clock away from the places that use it to force people who use this later to use `steady_clock` instead of `system_clock` (it doesn't really matter on these robots because we probably are never going to add timezone support in, but regardless it's probably better practice).

robot/control/modules/motion-control/RobotEstimator.cpp
```diff
@@ -0,0 +1,48 @@
...
+    K_cam = (Matrix<6, 3>() <<
+            0, 0, 0,
+            0, 0, 0,
+            0, 0, 0,
+            0, 0, 0,
+            0, 0, 0,
+            0, 0, 0
+            ).finished();
+    x.block<3, 1>(0, 0) = x0;
+    x.block<3, 1>(3, 0) = v0;
```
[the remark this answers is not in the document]
kylestach: Probably not...would you suggest to let these just be zero-initialized and then let them converge when they get in range of vision?

robot/control/modules/motion-control/RobotModel.cpp
kylestach: I think so. The `coriolis` matrix is the skew-symmetric representation of `omega x` (multiplication by `coriolis` is the same as the cross product with the angular velocity vector).

robot/control/modules/motion-control/RobotModel.hpp
```diff
@@ -0,0 +1,102 @@
...
+using Matrix = Eigen::Matrix<double, N, M>;
+
+template<int N>
+using Vector = Matrix<N, 1>;
+
+constexpr double to_radians(double degrees) {
+    return degrees * M_PI / 180;
+}
+
+constexpr double kLoopTime = 0.005;
```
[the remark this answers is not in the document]
kylestach: Yes. It's subject to change, though.

robot/control/modules/motion-control/MotionControl.hpp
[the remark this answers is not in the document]
JNeiger: I'm down, that sounds like a good enough reason to me

kylestach:
@JNeiger good catch, I had meant to use those everywhere but I guess I missed a few spots. the `block` code is just converting position and velocity to/from a 6-vector to 2 3-vectors, but I can look at cleaning it up.

robot/control/modules/motion-control/RobotModel.cpp
JNeiger: I guess what's throwing me off is that the usual place you see the coriolis effect is when you are find the acceleration of a point moving relative to frame B, which is moving relative to frame A. I don't think that's what we're doing here with the frame definitions?

robot/control/modules/motion-control/RobotEstimator.cpp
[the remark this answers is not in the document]
JNeiger: I think it's probably fine to just zero-initialize assuming they converge within a half second ish. In almost every single case that I can think of , we will have the robot turned on and in vision for at least a few seconds before moving.

robot/control/modules/motion-control/RobotModel.cpp
kylestach: Honestly I'm not sure I got the terminology right, but I think it applies? You've got the spinning reference frame of the robot and the inertial world frame, right?

kylestach:
Also, @Zolkin1 for reference: this code currently generates voltages, but it's super easy to drop that part off once we have torque control ready. I'm just assuming that we're going to want to run this so we can tune it before we've got torque control running.

robot/control/modules/motion-control/RobotModel.cpp
kylestach: FYI, what this actually does is captures the behavior of the robot when it has both linear and angular velocity (it experiences force perpendicular to its linear velocity and curves off to the side).

JNeiger: I believe the only extra thing is that the object has to have a non-zero velocity in both reference frames. In the non-inertial frame, the object will appear to have a secondary acceleration acting upon it. If you kick a ball and view it from the robot frame (with a non-zero angular velocity), you'll see the ball veer off to the left or right even though it has no other forces acting on it in the inertial frame.

If it has 0 velocity in one of the frames, it's probably a euler or centrifugal. Based on your edit, centrifugal might be a better term. It's like that sideways outwards force you feel in a car when you are turning, centripetal force is the one inwards

JNeiger: Centrifugal is -omega cross (omega cross r)
Euler is alpha cross r
Centripetal is omega cross (omega cross r)
Coriolis is 2 omega cross v
All have a extra inertia coefficient to convert accelerations to forces

PR #75, opened 2019-05-16
Wrapper for radio

JNeiger:
Wraps the radio driver to take the raw data used in firmware and send it to the base station (and vice versa).

I implemented a new interface for the radio to follow so that we can swap it out on the fly without actually changing code.

There's a new thing called micropackets for lack of a better name. It bundles the stuff in the packets together so it's a little cleaner to pass the data to the specific modules. Eg: All the velocity commands are related so we might as well pack them together into a single struct.

This probably doesn't actually compile yet. I still need to get the environment setup


guyfleeman:
Only thing I see that could be major is that we need to verify std::lock_guard has a valid implementation in the STM32F7 eABI.

JNeiger:
Merged into the super_loop branch for integration

PR #76, opened 2019-06-05
mTrain Firmware

JNeiger:


JNeiger:
Things left to do:
[ ] Debug motor error led's. I think the fpga response -> LED module isn't correct. Some LED's don't work either
[ ] Get kicker firmware fixed. Currently doesn't boot kickerboard after flash
[ ] Tune motion control
[ ] Swap to microsecond waits for everything instead of ms

Things I maybe should add:
[ ] Stall detection. Do we need this??

Things I won't try to do:
[ ] Neopixel. The mbed uses some assembly to get the timer right. I've found dma based examples online, but it's going to be a struggle and not that important

JNeiger:
Killing for comp branch

PR #78, opened 2019-07-04
Comp2019

JNeiger:
Real comp branch


robot/lib/Src/drivers/ISM43340.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+    isInit = !interruptin_read(dataReady);
+
+    if (isInit) {
+        printf("Could not initialize radio\r\n");
+        return;
+    }
+    
+
+    nCs = ISMConstants::CHIP_SELECT;
```
JNeiger: Everything below this should really be part of a config function

robot/lib/Src/drivers/ISM43340.cpp
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    nCs = ISMConstants::CHIP_DESELECT;
+
+    // Wait till data ready goes to 0
+    //while (dataReady.read() != 0);
+    while (currentState != ISMConstants::State::CommandAck);
+}
+
+uint32_t ISM43340::readFromSpi() {
```
JNeiger: Make sure radio is configured

robot/lib/Src/drivers/ISM43340.cpp
```diff
@@ -0,0 +1,430 @@
...
+ * and the ones given are partially wrong
+ * 
+ * Bump the super loop call rate until it just barely works
+ * Drop these delays until it just barely works
+ * Bump the super loop call rate again until it works
+ * repeat
+ * 
+ * Let it run for a minute or so to see if it breaks eventually
+ */
+void ISM43340::writeToSpi(uint8_t* command, int length) {
```
JNeiger: Make sure radio is configured

robot/lib/Src/drivers/KickerBoard.cpp
```diff
@@ -50,33 +47,33 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
     }
 
     //  Open binary file to write to AVR.
-    FILE* fp = fopen(_filename.c_str(), "r");
+    FILE* fp = fopen(proFilename.c_str(), "r");
 
     if (fp == nullptr) {
-        LOG(WARN, "Failed to open kicker binary, check path: '%s'",
-            _filename.c_str());
+        //LOG(WARN, "Failed to open kicker binary, check path: '%s'",
```
JNeiger: Either remove the logs, or implement them correctly

robot/lib/Src/drivers/KickerBoard.cpp
```diff
@@ -86,121 +83,136 @@ bool KickerBoard::flash(bool onlyIfDifferent, bool verbose) {
...
-    if (_stop_charging_commanded) {
-        _stop_charging_commanded = false;
+    command |= static_cast<uint8_t>(static_cast<float>(_kick_strength)/255 * 0xF) & KICK_POWER_MASK;
 
-        if (_is_charging) {
-            send_to_kicker(SET_CHARGE_CMD, OFF_ARG, nullptr);
-        }
-    }
+    // Transmit byte over to kicker
+    // Must wait 10 us such that the isr actually triggers
```
JNeiger: Fix comments

robot/lib/Src/drivers/MCP23017.cpp
```diff
@@ -1,11 +1,8 @@
...
+#include "drivers/MCP23017.hpp"
 
-#include "Logger.hpp"
-#include "Mbed.hpp"
-
-MCP23017::MCP23017(std::shared_ptr<SharedI2C> sharedI2C, int i2cAddress)
+MCP23017::MCP23017(std::shared_ptr<I2C> sharedI2C, int i2cAddress)
     : _i2c(sharedI2C), _i2cAddress(i2cAddress) {
-    _i2c.frequency(400000);
+    //_i2c.frequency(400000);
```
JNeiger: Remove comment

util/flash-ism43340
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
JNeiger: Should probably add some instructions to find the required bin file

robot/lib/Inc/drivers/ISM43340.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+    static const unsigned int RESET_DELAY = 500;
+
+    // Network config
+    static const std::string NETWORK_SSID = "rj-rc-field";
+    static const std::string NETWORK_PASSWORD = "r0b0jackets";
+
+    static const std::string ROUTER_IP = "172.16.1.1";
+
+    static const std::string BASE_STATION_IP = "172.16.1.48";//48//27//36
```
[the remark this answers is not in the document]
JNeiger: In the meantime we should swap it back to the desktop probably

convert.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
```
JNeiger: This file should have some comments describing an example use for the kicker and fpga

robot/conanfile.py
```diff
@@ -8,10 +8,12 @@ class RoboCupFirmwareConan(ConanFile):
     settings = {"build_type": None, "compiler": None, "arch": ["armv7hf"]}
     generators = "cmake"
 
-    requires = 'mTrain/[0.1.0]@evan/testing'
+    requires = (('mTrain/[0.1.6]@joe/testing'),
```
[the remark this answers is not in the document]
JNeiger: I'm going to need help pushing the latest mtrain version to bintray @petersonev

JNeiger:
This is all good I think. I can't test it, but I didn't change any logic so it should be fine

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+extern DebugInfo debugInfo;
+
+/**
+ * Edited sign function to always return -1 or 1
+ * 
+ * Returns 1 if val >= 0, -1 if val < 0
+ */
+template <typename T> int sgn(T val) {
+    return (T(0) <= val) - (val < T(0));
+}
```
kylestach: `std::signbit`

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+double limit(double val, double maxabs) {
+    if (val > maxabs) {
+        return maxabs;
+    } else if (val < -maxabs) {
+        return -maxabs;
+    } else {
+        return val;
+    }
+}
```
kylestach: `std::clamp`

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * Edited sign function to always return -1 or 1
+ * 
+ * Returns 1 if val >= 0, -1 if val < 0
+ */
+template <typename T> int sgn(T val) {
+    return (T(0) <= val) - (val < T(0));
+}
+
+template <typename T> T bounded(T val, T min, T max) {
```
kylestach: This should return a boolean, not a `T`.

kylestach: Although it's unused so really it should just be removed.

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+double ROBOT_RADIUS = 0.0794;
+
+void apply_wheel_force(const Eigen::Matrix<double, 4, 1> force, const Eigen::Matrix<double, 4, 1> speeds, Eigen::Matrix<double, 4, 1>& outputs) {
+    double CURRENT_PER_TORQUE = 1.0 / 25.1e-3;
+    double PHASE_RESISTANCE = 0.464;
+    for (int i = 0; i < 4; i++) {
+        double torque = force(i) * RobotModel::get().WheelRadius / 3.0;
+        double voltage = torque * CURRENT_PER_TORQUE * PHASE_RESISTANCE / 24.0;
+        double back_emf = (double) speeds(i) * RobotModel::get().SpeedToDutyCycle / 512.0;
+        outputs(i, 0) = 0.14 * voltage + back_emf * 0.7;
```
kylestach: We should put these constants as "real values" somewhere.

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+double JERK_LIMIT = 10.0;
+double TICK_TIME = 5e-3;
+double ROBOT_MASS_X = 6.35;
+double ROBOT_MASS_Y = 6.35;
+double ROBOT_MASS_H = 6.35 * 0.37;
+double ROBOT_RADIUS = 0.0794;
+
+void apply_wheel_force(const Eigen::Matrix<double, 4, 1> force, const Eigen::Matrix<double, 4, 1> speeds, Eigen::Matrix<double, 4, 1>& outputs) {
+    double CURRENT_PER_TORQUE = 1.0 / 25.1e-3;
+    double PHASE_RESISTANCE = 0.464;
```
kylestach: These should be pulled out to the top.

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+        return val;
+    }
+}
+
+double JERK_LIMIT = 10.0;
+double TICK_TIME = 5e-3;
+double ROBOT_MASS_X = 6.35;
+double ROBOT_MASS_Y = 6.35;
+double ROBOT_MASS_H = 6.35 * 0.37;
+double ROBOT_RADIUS = 0.0794;
```
kylestach: These should be constexpr and follow `kJerkLimit` convention.

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    WheelPrevTarget << 0, 0, 0, 0;
+}
+
+void RobotController::calculateBody(Eigen::Matrix<double, numStates, 1> pv,
+                                    Eigen::Matrix<double, numStates, 1> sp,
+                                    Eigen::Matrix<double, numWheels, 1>& outputs) {
+    if (std::abs(sp(0)) > 6.0) {
+        sp(0) = sgn(sp(0)) * 6.0;
+    }
```
kylestach: This can be replaced with `std::clamp`.

Also, 6 seems _really really really really_ high for sideways motion, right?

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    static Eigen::Matrix<double, numStates, 1> prev_linear_accel(0, 0, 0);
+
+    Eigen::Matrix<double, numStates, 1> linear_diff = linear_accel - prev_linear_accel;
+
+    linear_diff(0, 0) = limit(linear_diff(0, 0), JERK_LIMIT * TICK_TIME);
+
+    linear_diff(1, 0) = limit(linear_diff(1, 0), JERK_LIMIT * TICK_TIME);
+
+    linear_diff(2, 0) = limit(linear_diff(2, 0), JERK_LIMIT * TICK_TIME * 10);
```
kylestach: ```// TODO D E S T R O Y```

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+                                     Eigen::Matrix<double, numWheels, 1>& outputs) {
+    outputs = sp;
+    return;
+}
+
+bool RobotController::limitBodyAccel(const Eigen::Matrix<double, numStates, 1> finalTarget,
+                                     Eigen::Matrix<double, numStates, 1>& dampened) {
+    Eigen::Matrix<double, numStates, 1> delta = finalTarget - BodyPrevTarget;
+
+    double accel[numStates] = {maxForwardAccel, maxSideAccel, maxAngularAccel};
```
kylestach: If we change this to an Eigen::Vector3d we can use:

```cpp
dampened = finalTarget;
double multiplier = dampened.cwiseQuotient(maxAccel).cwiseAbs().cwiseMax();
if (multiplier > 1) {
    dampened /= multiplier;
}
```

robot/control/Src/motion-control/RobotController.cpp
```diff
@@ -0,0 +1,186 @@
...
+            limited = true;
+        }
+
+        if (delta(i, 0) < -dt*maxWheelAccel) {
+            delta *= -dt*maxWheelAccel / delta(i, 0);
+            limited = true;
+        }
+    }
+
+    dampened = delta + WheelPrevTarget;
```
kylestach: Honestly this could all just be a function templated over the size of the input matrix, in which case it would be the same as the above.

robot/control/Src/motion-control/RobotEstimator.cpp
```diff
@@ -0,0 +1,68 @@
...
+    P.setIdentity();
+    P *= initCovariance;
+
+    I.setIdentity();
+
+    // Assume no motion
+    x_hat << 0, 0, 0;
+}
+
+// 1.3 ms
```
kylestach: Wait, what? Shouldn't the estimator be running at 200Hz with the rest of the MotionControl stack?

robot/control/Src/motion-control/RobotEstimator.cpp
```diff
@@ -0,0 +1,68 @@
...
+void RobotEstimator::predict(Eigen::Matrix<double, numInputs, 1> u) {
+    // x_hat = F*x_hat + B*last command
+    // P = F*P*F' + Q
+    x_hat = F*x_hat + B*u;
+    P = F*P*F.transpose() + Q;
+}
+
+// 7 ms
```
kylestach: I would think this would be the same as the above.

kylestach: (Considering we do not use vision in the KF at this point)

robot/control/Src/radio/RadioLink.cpp
```diff
@@ -0,0 +1,85 @@
...
+    status->uid             = robotID.robotID;
+    status->battVoltage     = batteryVoltage.rawVoltage;
+    status->motorErrors     = motorErrors;
+    status->ballSenseStatus = static_cast<uint8_t>(kickerInfo.ballSenseTriggered);
+    status->kickStatus      = static_cast<uint8_t>(kickerInfo.kickerCharged);
+    status->kickHealthy     = static_cast<uint8_t>(kickerInfo.kickerHasError);
+    status->fpgaStatus      = static_cast<uint8_t>(fpgaStatus.FPGAHasError);
+
+    for (int i = 0; i < 18; i++)
+        status->encDeltas[i] = debugInfo.val[i];
```
kylestach: Should probably rename this field.

robot/control/main.cpp
```diff
@@ -1,31 +1,195 @@
...
+            uint32_t currentTime = DWT_GetTick();
+
+            // Check if we should run the module
+            //      It's time to run it and
+            //      we have enough time in the super loop
+            //
+            // Subtraction allows for rollover compensation
+            // then convertion to signed allows simple comparison
+            if (static_cast<int32_t>(currentTime - module.nextRunTime) >= 0 &&
+                static_cast<int32_t>(loopEndTime - currentTime) >= module.moduleRunTime) {
```
kylestach: :cry:

robot/lib/Inc/drivers/GenericRadio.hpp
```diff
@@ -0,0 +1,37 @@
...
+     * Blocking call to send X number of bytes from `data` over the radio
+     * 
+     * @param data raw array of data to send over the radio
+     * @param numBytes number of bytes in the array to send
+     * 
+     * @return number of bytes sent
+     * 
+     * @note It is assumed that `data` is at least X bytes long
+     */
+    virtual unsigned int send(const uint8_t* data, const unsigned int numBytes) = 0;
```
kylestach: Why do we need dynamic dispatch?

We should be explicitly minimizing use of `new` in firmware, and dynamic dispatch encourages people to do that. And we _really_ don't need it here IMO.

robot/lib/Inc/drivers/GenericRadio.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Blocking call to write up to `maxNumBytes` into `data` from the radio
```
kylestach: > Blocking

:cry:

robot/lib/Inc/drivers/Internal/helper_3dmath.h
```diff
@@ -32,6 +32,8 @@ THE SOFTWARE.
 #ifndef _HELPER_3DMATH_H_
 #define _HELPER_3DMATH_H_
 
+#include <cmath>
+
 class Quaternion {
```
kylestach: FWIW Eigen supports quaternions. Might be nice to dump this.

kylestach: Actually, why do we even have this in the first place?

robot/lib/Src/drivers/ISM43340.cpp
```diff
@@ -0,0 +1,424 @@
+#include "drivers/ISM43340.hpp"
+#include "delay.h"
+#include <cstring>
+#include "interrupt_in.h"
+
+volatile ISMConstants::State currentState;
+
+// Callback on every state change on data ready
+void dataReady_cb() {
+    // lmao
```
kylestach: lmao

robot/lib/Src/drivers/MCP23017.cpp
```diff
@@ -24,14 +20,13 @@ void MCP23017::reset() {
...
+    std::vector<uint8_t> buffer{static_cast<uint8_t>(data & 0xff),
+                                static_cast<uint8_t>(data >> 8)};
+    _i2c->transmit(_i2cAddress, regAddress, buffer);
 }
 
 uint16_t MCP23017::readRegister(MCP23017::Register regAddress) {
-    char buffer[2];
-    _i2c.write(_i2cAddress, (char*)&regAddress);
-    _i2c.read(_i2cAddress, buffer, 2);
+    std::vector<uint8_t> buffer = _i2c->receive(_i2cAddress, regAddress, 2);
```
kylestach: Putting it here because I don't have a better place for it (probably in the mtrain repo but I'm lazy).

`<kyle dynamic allocation rant>`

I am not a fan of this. This heap-allocates a 2-byte value. Repeatedly. To read a register. AND the size is bounded at compile time (I assume) because you are passing in 2.

I would _strongly_ prefer to at least have a second API for receive that allows you to pass in a pointer to a buffer and a size. Same with transmit above. This is overkill and physically pains me.

`</kyle dynamic allocation rant>`

robot/control/Src/modules/RadioModule.cpp
```diff
@@ -0,0 +1,61 @@
...
+                         KickerCommand *const kickerCommand,
+                         MotionCommand *const motionCommand,
+                         RadioError *const radioError)
+    : batteryVoltage(batteryVoltage), fpgaStatus(fpgaStatus),
+      kickerInfo(kickerInfo), robotID(robotID),
+      kickerCommand(kickerCommand), motionCommand(motionCommand),
+      radioError(radioError), link(),
+      secondRadioCS(RADIO_R1_CS) {
+
+    secondRadioCS = 1;
```
kylestach: One more comment I forgot to make. Can we _please_ just use `.write()` instead of `operator=` for setting DIO values? This is so confusing - I had to read this code several times before I realized that `secondRadioCS` isn't an integer.

I also put an issue in mtrain-firmware to remove this API. IMO it does nothing but create confusion.

robot/control/Src/motion-control/RobotEstimator.cpp
JNeiger: We were trying to figure out what was taking so long in this. Iirc we didn't have optimizations on so these functions actually took. It's fixed now

robot/lib/Inc/drivers/GenericRadio.hpp
JNeiger: This was just an easy way to abstract out the direct radio driver from everything else. It's very annoying when we swap devices every other year and nobody actually removes and replaces the radio code in a clean way to fit with a specific device.

My hope with this was to enforce an API such that it's not a problem (given it's not async). It provides a clear line between what should be device specific code and what should be application code. In my mind, that clear separation is more useful than the overhead this creates.

kylestach: That's fair. My point was more that I don't want to encourage people to dynamically allocate things that could be static. As long as we avoid that I'm fine.

robot/control/Src/motion-control/RobotEstimator.cpp
kylestach: Yikes. Okay. As a side note we should really probably be using `float` instead of `double` everywhere, 64-bit floating point is probably overkill for what we're doing and it's almost certainly significantly slower considering we have native hardfp32 support now. Thoughts?

JNeiger: The entire controller / estimator is under 300 us now. We could switch if want to based on hardware support, but it's starting to get to the point it doesn't matter. IO is still the bulk of the transactions.
I'm curious to see just how much faster it is if you swapped to that.

kylestach: That 300us is just computations and nothing else, right? Seems a little bit high for what effectively comes down to a few matrix multiplications and an inverse.

JNeiger: Scratch that, just checked, it's 133 us (all the module hpps have the estimates), which at a 200 MHz doesn't actually feel that far off. Try it and see

kylestach: Ah. That seems fairly reasonable, then - still a little higher than I'd expect but I'll look into it later if it ends up being a concern.

kylestach:
I've cleaned up my stuff on the motion control side of things.

PR #83, opened 2019-09-25
Firmware Docs

JNeiger:
Still have a little bit to do, but I kinda want to keep this same style and detail.

@mmaisonneuve3 Do a double check before I keep going


CollinAvidano:
Merged as part of #92

PR #86, opened 2019-10-23
Comp2019

Arvind-Srinivasan:
If we are doing major development off of this branch, maybe it should be part of master? Thoughts?


kylestach:
Seconded.

Arvind-Srinivasan:
I'm down to merge, any opposition?

PR #90, opened 2019-11-18
RTOS super loop implementation

kylestach:
This is an initial implementation of RTOS on mTrain that just moves the old super-loop code into its own task. It makes no attempt to actually use any features of RTOS.


robot/control/main.cpp
```diff
@@ -145,51 +195,12 @@ int main() {
...
 
-    while (true) {
-        // Must do all timing in systick
-        // Base time we use is in systick, and if you convert from systick
-        // to us and try to compare timepoints, the value will "overflow" at
-        // some random point which messes up the `Current - Start` magic
-        // that automatically deals with delta time using the math behind overflows
-        uint32_t loopStartTime = DWT_GetTick();
-        uint32_t loopEndTime = DWT_GetTick() + SUPER_LOOP_PERIOD*DWT_SysTick_To_us();
+    osThreadDef(SuperLoopTask, superLoopTask, osPriorityAboveNormal, 1, 1 << 15);
```
JNeiger: Do you have any docs for this call? I'm only seeing the 4 arg version in the keil cmsis docs here. https://www.keil.com/pack/doc/CMSIS/RTOS/html/group__CMSIS__RTOS__ThreadMgmt.html#gaee93d929beb350f16e5cc7fa602e229f

robot/control/main.cpp
```diff
@@ -72,6 +73,58 @@ KickerInfo kickerInfo;
...
+            DWT_Delay_Sys(SUPER_LOOP_PERIOD*DWT_SysTick_To_us() - elapsed);
+        } else {
+            //printf("WARNING: Overran super loop time\r\n");
+            led->missedSuperLoop();
+        }
+    }
+}
+
 
 int main() {
```
JNeiger: You need a 'osKernalInitialize' call per https://www.keil.com/pack/doc/CMSIS/RTOS/html/group__CMSIS__RTOS__KernelCtrl.html

robot/control/main.cpp
```diff
@@ -145,51 +195,12 @@ int main() {
...
-    while (true) {
-        // Must do all timing in systick
-        // Base time we use is in systick, and if you convert from systick
-        // to us and try to compare timepoints, the value will "overflow" at
-        // some random point which messes up the `Current - Start` magic
-        // that automatically deals with delta time using the math behind overflows
-        uint32_t loopStartTime = DWT_GetTick();
-        uint32_t loopEndTime = DWT_GetTick() + SUPER_LOOP_PERIOD*DWT_SysTick_To_us();
+    osThreadDef(SuperLoopTask, superLoopTask, osPriorityAboveNormal, 1, 1 << 15);
+    osThreadCreate(osThread(SuperLoopTask), NULL);
```
JNeiger: I assume we don't really care if threads fail to init. The entire system will be dying anyways

robot/control/main.cpp
```diff
@@ -145,51 +195,12 @@ int main() {
...
-        int32_t elapsed = DWT_GetTick() - loopStartTime;
-        if (elapsed < static_cast<int32_t>(SUPER_LOOP_PERIOD*DWT_SysTick_To_us())) {
-            DWT_Delay_Sys(SUPER_LOOP_PERIOD*DWT_SysTick_To_us() - elapsed);
-        } else {
-            //printf("WARNING: Overran super loop time\r\n");
-            led.missedSuperLoop();
-        }
-        
-    }
+    for (;;);
```
JNeiger: I always like throwing a yield in here

kylestach: This should be unreachable.

robot/control/main.cpp
kylestach: Yeah, this whole thing will be getting the axe. This is just to get RTOS stuff up-and-running ASAP so we can verify things don't break.

robot/control/main.cpp
kylestach: ACK

robot/control/main.cpp
kylestach: Nope. I looked at the header file - the version that we have is:

```cpp
#define osThreadDef(name, thread, priority, instances, stacksz)
```

robot/control/main.cpp
kylestach: Yeah, I'm okay with not handling errors for now. This will be getting the axe soon, it's just to make sure RTOS doesn't break anything horrifically and to keep the process moving.

robot/control/main.cpp
JNeiger: How so? osKernalStart returns and continues execution in main base on my reading of the following note.

Note
When the CMSIS-RTOS starts thread execution with the function main this function resumes thread switching. The main thread will continue executing after osKernelStart.

robot/control/main.cpp
kylestach: As a side note, I mentioned on slack that it might be a good idea to use FreeRTOS APIs instead of the CMSIS wrapper layer - they are certainly better documented, which might be a point in favor.

robot/control/main.cpp
kylestach: Huh. That is different from the documentation that I've seen...I think that note may be referring to when you actually have the initialization sequence in a Task, i.e. main is a task, which we're not doing.

From https://www.keil.com/pack/doc/CMSIS/RTOS2/html/theory_of_operation.html:
> Start the RTOS scheduler using osKernelStart which also configures the system tick timer and initializes RTOS specific interrupts. This function does not return in case of successful execution. Therefore, any application code after osKernelStart will not be executed.

robot/control/main.cpp
kylestach: Uhh...this is actually unimplemented for our port. I'm very confused. It also doesn't seem to be necessary somehow?

kylestach:
```cpp
/**
* @brief  Initialize the RTOS Kernel for creating objects.
* @retval status code that indicates the execution status of the function.
* @note   MUST REMAIN UNCHANGED: \b osKernelInitialize shall be consistent in every CMSIS-RTOS.
*/
osStatus osKernelInitialize (void);
```

I am not impressed by this CMSIS-RTOS layer.

robot/control/main.cpp
JNeiger: Oof. I was looking at the RTOS1 docs not RTOS2

JNeiger: Based on the `osStatus` return type on `osKernalInitialize`, I think this actually might be RTOS v1. In v2 the return type is now `osStatus_t`. `osThreadDef` uses a param struct and `osThreadCreate` is called `osThreadNew` in v2.

[Changelog](https://www.keil.com/pack/doc/CMSIS/RTOS2/html/os2MigrationFunctions.html)

In the [migration guide](https://www.keil.com/pack/doc/CMSIS/RTOS2/html/os2MigrationGuide.html), it mentions that main is actually it's own thread, which might make that note applicable again. In v2 it specifies that it will not return, but v1 does not have the same description so I think the behavior is different.

Also, it very much make sounds like `osKernelInitialize` and `osKernelStart` aren't fully required, but highly suggested if you need to set things up without switching threads accidentally. Since the [default define](https://www.keil.com/pack/doc/CMSIS/RTOS2/html/os2MigrationGuide.html) is for main thread to be on, calling `osKernelInitialize` pauses thread switching until `osKernelStart` is called again.

kylestach:
Alright I dropped CMSIS primitives. Documentation for regular FreeRTOS things is much better.

robot/control/main.cpp
```diff
@@ -145,51 +196,16 @@ int main() {
...
-    while (true) {
-        // Must do all timing in systick
-        // Base time we use is in systick, and if you convert from systick
-        // to us and try to compare timepoints, the value will "overflow" at
-        // some random point which messes up the `Current - Start` magic
-        // that automatically deals with delta time using the math behind overflows
-        uint32_t loopStartTime = DWT_GetTick();
-        uint32_t loopEndTime = DWT_GetTick() + SUPER_LOOP_PERIOD*DWT_SysTick_To_us();
+    TaskHandle_t superLoopHandle;
+    xTaskCreate(superLoopTask, "Super Loop", 1 << 15, nullptr, 1, &superLoopHandle);
```
JNeiger: `size_t` is larger than `1 << 15` right?

kylestach: Yeah, size_t should be 32 bits.

kylestach:
I'll merge this on Sunday, once I've confirmed that it doesn't break anything.

kylestach:
In that latest commit I had to add two copies of `rtos_hooks.c`, one for each target. That's because for whatever reason, it wasn't finding symbols (`vApplicationGetIdleTaskMemory` and `vApplicationGetTimerTaskMemory`) properly with the file in `firm-lib`, even though `nm` clearly shows that the aforementioned symbols do in fact exist in `firm-lib.a`. If someone with a better understanding of CMake wants to debug it that's fine by me, but I am plain confused and am honestly fine with having two copies of the file lying around as a workaround.

kylestach:
Tested, this works.

kylestach:
This has been tested and works correctly. I'm merging it.

PR #91, opened 2019-11-24
Add Armv7hf and remotes when setting up firmware.

CollinAvidano:
Small addition to the ubuntu setup script which adds the robojackets conan remotes and a copy of armv7hf profile stored locally. Should be at a url in the future for both this and mtrain repo. Also removed submodule update on external/ which no longer exists.


kylestach:
I agree that the remote needs to be added. However the armv7hf file already exists in robot/, and it should be found automatically by Conan.

CollinAvidano:
Ignore this PR merging things into the branch support-update

PR #95, opened 2020-02-02
RTOS rewrite

kylestach:
Rewrite the entire code base to use RTOS.

Design document: https://docs.google.com/document/d/1767ShGdEDZaqAaYE4M8tanvGCliVxE0DQHzV9dSy7es/edit#

I'm so sorry for the size of this PR.


robot/control/Src/modules/RadioModule.cpp
```diff
@@ -17,45 +18,69 @@ RadioModule::RadioModule(BatteryVoltage *const batteryVoltage,
...
+    {
+        auto batteryVoltageLock = batteryVoltage.lock();
+        auto fpgaStatusLock = fpgaStatus.lock();
+        auto robotIDLock = robotID.lock();
+        auto kickerInfoLock = kickerInfo.lock();
+
+        // Just check to see if our robot id is valid
+        // That way we don't conflict with other robots on the network
+        // that are working
+//        if (batteryVoltageLock->isValid && fpgaStatusLock->isValid && robotIDLock->isValid) {
```
JNeiger: Depending on if the robotID is read before this or not, it's possible that this robot will connect to soccer as `robotID=0` for a single frame then swaps to the correct robotID. I believe soccer starts to act weird as more MAC<->RobotID combinations connect to soccer. The robots start to be only intermittently connected and you have to restart soccer to fix

robot/control/Inc/iodefs.h
```diff
@@ -12,14 +12,16 @@
 #define RADIO_R1_INT  p18
 #define RADIO_GLB_RST p19
 
-#define FPGA_KICKER_SPI_BUS SpiBus::SpiBus3
+#define FPGA_SPI_BUS SpiBus::SpiBus3
```
CollinAvidano: Thank you so much

kylestach:
This is _almost_ ready to be merged. I still need to verify that it doesn't break kicker functionality.

robot/control/Src/modules/RadioModule.cpp
kylestach: Fixed, this was only temporarily disabled.

PR #96, opened 2020-02-09
Motor test script

matthewawhite:
Have fun Arvind, I'll let you add in all the documentation to the code.

### DIFF
--- a/robot/control/Src/modules/RotaryDialModule.cpp
+++ b/robot/control/Src/modules/RotaryDialModule.cpp
@@ -19,3 +19,9 @@ void RotaryDialModule::entry(void) {
     robotID->isValid = true;
     robotID->lastUpdate = HAL_GetTick();
 }
+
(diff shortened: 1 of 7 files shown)


robot/control/Src/modules/RotaryDialModule.cpp
```diff
@@ -18,4 +18,10 @@ void RotaryDialModule::entry(void) {
         robotID->robotID = dial.read();
     robotID->isValid = true;
     robotID->lastUpdate = HAL_GetTick();
+}
+
+void RotaryDialModule::motorTestEntry(void) {
```
JNeiger: You can actually change the other entry to continuously get the dial value. It only happened once because when I was debugging the I2C IMU issue, I lost manual control because the ID would instantly change to 0 when the I2C bus died. If that isn't a concern anymore, it's fine to change it back

PR #97, opened 2020-02-19
Icm 20948 driver working test

Tohwx:


robot/control/main.cpp
```diff
@@ -119,6 +119,8 @@ int main() {
     //                     &kickerCommand,
     //                     &kickerInfo);
 
+    led.radioInitialized();
```
kylestach: Revert this file, nothing should change here.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    constexpr uint8_t USER_CTRL     = 0x03;
+    constexpr uint8_t LP_CONFIG     = 0x05; 
+    constexpr uint8_t PWR_MGMT_1    = 0x06;
+    constexpr uint8_t PWR_MGMT_2    = 0x07;
+
+    constexpr uint8_t GYRO_Z_OUT_H  = 0x37;
+    constexpr uint8_t GYRO_Z_OUT_L  = 0x38;
+
+    constexpr uint8_t FIFO_MODE     = 0x69;
+    //constexpr uint8_t
```
kylestach: Cleanup: delete this comment and the similar one below.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+void ICM20948::chip_select(bool cs_state) {
+    if (cs_state) {
+        nCs.write(0);
+    } else {
+        nCs.write(1);
+    }
+}
+
+void ICM20948::write_register(uint8_t bank, uint8_t address, uint8_t value) {
+    chip_select(true);
```
kylestach: Only set bank select if it's changed from the last one set.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    imuSPI->transmit(address);
+    imuSPI->transmit(value);
+    chip_select(false);
+}
+
+
+uint8_t ICM20948::read_register(uint8_t bank, uint8_t address) { 
+    chip_select(true);
+    imuSPI->transmit(Registers::REG_BANK_SEL);
+    imuSPI->transmit(bank);
```
kylestach: Same bank select comment as above.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+*/
+
+void ICM20948::chip_select(bool cs_state) {
+    if (cs_state) {
+        nCs.write(0);
+    } else {
+        nCs.write(1);
+    }
```
kylestach: This can be simplified to `nCs.write(!cs_state);`

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    chip_select(was_cs_active);
+}
+
+void ICM20948::chip_select(bool active) {
+    if (currently_active != active) {
+        currently_active = active;
+        chip_select_pin.write(active);
+    }
+}
+*/
```
kylestach: The commented-out code here can be deleted.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    uint8_t whoami = read_register(0, Registers::WHO_AM_I);
+    bool failed_init = false;
+
+    if (whoami != WHOAMI_VAL) {
+        printf("Failed to connect to IMU.");
+        failed_init = true;
+    }
+
+    // Reset
+    write_register(0, Registers::PWR_MGMT_1, 0xC1);
```
kylestach: These magic constants like `0xC1` aren't super helpful. Can you break them out into their component parts? See the ICM42605 driver for examples:
```cpp

    write_register(
            Registers::GYRO_CONFIG1,
            Registers::GyroTempFiltBW::HZ_4000 << 5
            | Registers::GyroFiltOrd::ORD_1 << 2
            | Registers::GyroFiltOrd::ORD_3); 
```

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    write_register(0, Registers::USER_CTRL, 0x10);
+
+    // Turn FIFO off
+    write_register(0, Registers::FIFO_MODE, 0x00);
+
+    // Enable gyro
+    write_register(0, Registers::PWR_MGMT_2, 0x00);
+    
+    // Configure gyro
+    write_register(2, Registers::GYRO_CONFIG_1, 0x04);
```
kylestach: I can't remember, does this set 1000dps mode? Make sure it does.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    HAL_Delay(1500);
+}
+
+bool ICM20948::initialize() {
+    // chip_select(true);
+
+    // SPI pulise to clear pins. 
+    imuSPI->transmit(0);    
+    
+    HAL_Delay(100);
```
kylestach: Remove this delay.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+ICM20948::ICM20948(std::shared_ptr<SPI> imuSPI, PinName cs_pin) 
+    : imuSPI(std::move(imuSPI)),
+      nCs(cs_pin) {
+    //chip_select_pin.write(false);
+    nCs = 1;
+    HAL_Delay(1500);
+}
+
+bool ICM20948::initialize() {
+    // chip_select(true);
```
kylestach: Remove.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+    
+    HAL_Delay(100);
+
+    uint8_t whoami = read_register(0, Registers::WHO_AM_I);
+    bool failed_init = false;
+
+    if (whoami != WHOAMI_VAL) {
+        printf("Failed to connect to IMU.");
+        failed_init = true;
+    }
```
kylestach: I would retry connection a few times before giving up.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    // Reset
+    write_register(0, Registers::PWR_MGMT_1, 0xC1);
+
+    HAL_Delay(100);
+
+    // Check whoami again
+    whoami = read_register(0, Registers::WHO_AM_I);
+
+    if (whoami != WHOAMI_VAL && failed_init) {
```
kylestach: This should be `||`, right? Otherwise this condition will be always false (because `failed_init` is only `true` whenever the above condition has already returned from the function).

Also, this should _definitely_ retry until it works properly. The `HAL_Delay` isn't a very good way of timing the gyro's boot time.

robot/lib/Src/drivers/ICM42605.cpp
```diff
@@ -0,0 +1,258 @@
+#include "drivers/ICM42605.hpp"
```
kylestach: This isn't part of this PR and shouldn't be included.

kylestach:
Also, please remember to delete your .swo files. They shouldn't get checked into the repository.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    uint16_t hi = read_register(0, Registers::GYRO_Z_OUT_H),
+             lo = read_register(0, Registers::GYRO_Z_OUT_L);
+
+
+    // Reinterpret the bits as a signed 16-bit integer
+    int16_t total = hi << 8 | lo;
+
+    // +-1000dps max, for 16-bit signed integers, comes out to 32.8lsb/deg.
+    const float lsbToDps = 1000.0f / 32768.f;
```
JNeiger: The data sheet almost always describes the scale factor to go from LSB to deg/sec. See table 1 in section 3.1. I have seen some devices in which the output of the gyro doesn't actually fill all 16 bits resulting in super incorrect numbers. In this case it's ok, but it's good to keep it in mind for the future.  The conversion is done following the instructions found in 8.25.

image
image

Also, these constants can be constexpr. It doesn't make too much of difference in this case, but it does the calculations at compile time instead of possibly during runtime.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+    while (whoami != WHOAMI_VAL) {
+        printf("Failed to connect to IMU.");
+        whoami = read_register(Registers::WHO_AM_I);
+    }
+    
+    // Reset
+    write_register( 
+        Registers::PWR_MGMT_1,
+        Registers::PWR_MGMT::DEVICE_RESET << 7
+        | Registers::PWR_MGMT::SLEEP << 6
```
JNeiger: This puts the device to sleep

robot/lib/Src/drivers/ICM20948.cpp
kylestach: This matches the data sheet (and the data sheet actually rounds in an undesirable way), so no worries in this case - I think it's preferable as-is.

I mean, this will almost certainly be optimized out at compile-time regardless. But yeah, it never hurts to make it constexpr.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+    
+    while (whoami != WHOAMI_VAL) {
+        printf("Failed to connect to IMU.");
+        whoami = read_register(Registers::WHO_AM_I);
+    }
+    
+    // Reset
+    write_register( 
+        Registers::PWR_MGMT_1,
+        Registers::PWR_MGMT::DEVICE_RESET << 7
```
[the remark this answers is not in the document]
kylestach: Yeah, I think that would be a good idea.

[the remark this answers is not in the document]
kylestach: Yes. It would probably cleaner to only write `DEVICE_RESET` here.

robot/lib/Src/drivers/ICM20948.cpp
kylestach: Doesn't really matter considering it's a reboot, but yes.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    HAL_Delay(100);
+
+    // Check whoami again
+    whoami = read_register(0, Registers::WHO_AM_I);
+    
+    while (whoami != WHOAMI_VAL) {
+        printf("Failed to connect to IMU.");
+        whoami = read_register(Registers::WHO_AM_I);
+    }
```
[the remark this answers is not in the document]
kylestach: Yeah, this shouldn't really be a "failure" so much as a "waiting for IMU to reboot".

It should also very much not be a spin loop, because the IMU runs at a super high priority and this will choke out other threads. Just add a `vTaskDelay` in.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+    
+    // Reset
+    write_register( 
+        Registers::PWR_MGMT_1,
+        Registers::PWR_MGMT::DEVICE_RESET << 7
+        | Registers::PWR_MGMT::SLEEP << 6
+        | Registers::PWR_MGMT::CLKSEL);
+
+    HAL_Delay(100);
```
kylestach: This needs to be a `vTaskDelay`.

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+        Registers::GYRO_FS_SEL::DPS_1000 << 2);
+    
+    
+    printf("Initialization Complete!");
+}
+
+float ICM20948::gyro_z() {
+
+    uint16_t hi = read_register(0, Registers::GYRO_Z_OUT_H),
+             lo = read_register(0, Registers::GYRO_Z_OUT_L);
```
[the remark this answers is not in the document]
kylestach: Good catch, we were focused on just getting something that worked quickly.

@Tohwx, I'd add a `burstRead` method that properly reads the number of bytes required given a start register.

robot/lib/Src/drivers/ICM20948.cpp
kylestach: Again, this.

Just record the latest bank selected (remember to reset it when resetting the device). This is going to cause a lot of extra SPI writes as-is (not the end of the world, just a little bit overkill).

robot/control/main.cpp
```diff
@@ -119,6 +119,7 @@ int main() {
     //                     &kickerCommand,
```
kylestach: Undo changes to this file.

robot/lib/Src/drivers/ICM42605.cpp
```diff
@@ -0,0 +1,257 @@
+#include <cmath>
```
kylestach: Delete this file, it's not part of this PR.

robot/lib/Src/drivers/ICM20948.cpp
JNeiger: Due to how the chip is made, I don't think we can actually say if the data sheet rounds the scale factor number or the full-range number. Either way, the +-3% tolerance on the scale factor greatly outweighs this rounding anyways so it's not a big deal at all.

I was mostly throwing it out there for info that this exists in the data sheet and can easily cause major problems. I think two years ago someone did this exact same calculation and it was a few orders of magnitude off resulting it horribly incorrect numbers

kylestach:
@Tohwx bump on this. Let's aim to get this merged in over the next week, please try to address any remaining projects.

Arvind-Srinivasan:
@kylestach can this PR be closed in favor of [this PR](https://github.com/RoboJackets/robocup-firmware/pull/111)

robot/lib/Src/drivers/ICM20948.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }; 
+    
+};
+
+ICM20948::ICM20948(std::shared_ptr<SPI> imuSPI, PinName cs_pin) 
+    : imuSPI(std::move(imuSPI)),
+      nCs(cs_pin) {
+    //chip_select_pin.write(false);
+    nCs = 1;
+    HAL_Delay(1500);
```
kylestach, line 55:
```suggestion
    vTaskDelay(1500);
```

robot/lib/Src/drivers/ICM20948.cpp
kylestach, line 76:
```suggestion
    vTaskDelay(100);
```

PR #98, opened 2020-02-23
Radio 3.2

kylestach:
Changes for radio 3.2 support


robot/lib/Src/drivers/ISM43340.cpp
```diff
@@ -294,30 +294,38 @@ void ISM43340::pingRouter() {
...
+        vTaskDelay(ISMConstants::RESET_DELAY);
+        nReset = ISMConstants::RESET_TURN_ON;
+        vTaskDelay(ISMConstants::RESET_DELAY);
+
+        radioSPI->frequency(ISMConstants::SPI_FREQ);
+
+        int counter = 0;
+        // Wait for device to turn on
+        for (int counter = 0; counter < 1500; counter++) {
+            vTaskDelay(10);
```
JNeiger: How long is this wait in wall time? It's 1500 * 10 * 5 if it fails which sounds like a long time. It also may not matter since radio is essential to operation

robot/lib/Src/drivers/ISM43340.cpp
```diff
@@ -294,30 +294,38 @@ void ISM43340::pingRouter() {
...
-    vTaskDelay(ISMConstants::RESET_DELAY);
-    nReset = ISMConstants::RESET_TURN_ON;
-    vTaskDelay(ISMConstants::RESET_DELAY);
+    for(int i = 0; i < 5; i++) {
+        nReset = ISMConstants::RESET_TURN_OFF;
+        vTaskDelay(ISMConstants::RESET_DELAY);
+        nReset = ISMConstants::RESET_TURN_ON;
+        vTaskDelay(ISMConstants::RESET_DELAY);
+
+        radioSPI->frequency(ISMConstants::SPI_FREQ);
```
JNeiger: You probably don't need to do this every try, but it doesn't hurt

robot/lib/Src/drivers/ISM43340.cpp
kylestach: Yeah this is an absurdly long time. It's not blocking other init processes because of RTOS though, so it is fine.

PR #105, opened 2020-05-01
AutoGen Docs

CollinAvidano:
Did a few things. Set up a workflow in github actions to autogen documentation which also required setting up a dockerfile for Firmware in the process and making minor tweaks to the setup script. I feel like a lot of the config for CI was really jank so please give feed back and ask questions.
I would really like some feed back as to what branches and events this workflow should run on as well as the job I am about to add as another PR that is for build testing.


.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
+name: CI
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master, develop ]
+  pull_request:
+    branches: [ master, develop ]
```
jgkamat, line 8: If this is purely to generate docs, I don't think it should run on pull requests, right? That way a "bad pr's docs" won't overwrite the published version. However, to test this pull request in particular, it could be useful.

Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Use phusion/baseimage if problems arise
+FROM ubuntu:18.04
+MAINTAINER Collin Avidano collin.avidano@gmail.com
+
+# Setup apt to be happy with no console input
+ENV DEBIAN_FRONTEND noninteractive
+
+# setup apt tools and other goodies we want
+# TODO Double check these later for which are really necessary
+RUN apt-get update --fix-missing && apt-get -y install udev git nano vim software-properties-common sudo apt-utils wget curl htop iputils-ping less bsdmainutils scons screen && apt-get clean
```
jgkamat, line 10: you probably don't need scons, screen, htop, nano, vim. However it's nice to have them for debugging in case things break, it's up to you whether you want to keep it or not.

.github/workflows/main.yml
CollinAvidano: I agree. I wasnt thinking. Arguably it really should just be pushes to master or develop (even though master is protected) and merges to master.

Dockerfile
CollinAvidano: Thank you. Will probably keep screen, and vim as this image is also so members can use it for dev if they want.

CollinAvidano:
Also wanted to let you all know I will probably close this PR without merging and you should probably focus on #106 as it was branched of this. I thought setting up a build test may take a bit but it took about 20 mins.

.github/workflows/main.yml
CollinAvidano: Just learned merges implictly cause a push even so just push to master or develop should be fine. Additionally the update to "docs" dont occur on the branch pushed. They occur TO the gh-pages branch based off the source in the new merged version of master or develop.

Dockerfile
jgkamat: Ah, if this has the possibility of being used for dev (I was thinking this one is just for CI), then all of these except scons might be useful. Scons is the old build system which I hope we are no longer using at all.

CollinAvidano:
Closed in favor of #106

PR #108, opened 2020-05-17
Removing circle CI resources

CollinAvidano:
Removing the couple of files left around from the old circle CI build system let me know if there are more.


CollinAvidano:
Also should I remove the old docker file Jay wrote. Like the one currently in use is literally a copy paste of it.

PR #109, opened 2020-06-11
More Documentation

tcontis:
Added some documentation, especially for modules, MicroPackets, and RobotEstimator, and converted some existing documentation to be more Doxygen-friendly

I hope it makes the code more accessible to people in the future, whether they look at Doxygen or look through the files themselves.

### DIFF
--- a/robot/control/Inc/MicroPackets.hpp
+++ b/robot/control/Inc/MicroPackets.hpp
@@ -23,7 +23,7 @@ struct MotionCommand {
     float bodyYVel;       /**< y-component of robot's linear velocity (m/s) */
     float bodyWVel;       /**< Angular velocity of robot about z axis (rad/s) */
 
-    uint8_t dribbler;     /**< 7-bit encoding of dribble rotation (0: no spin - 128: max spin)
+    uint8_t dribbler;     /**< 7-bit encoding of dribbler rotation (0: no spin - 128: max spin)
                           @note **This is limited from 0-128** */
 };
 
@@ -174,4 +174,3 @@ struct KickerInfo {
 struct DebugInfo {
     int16_t val[18]; /**< Array of integers of debug info for soccer */
 };
-
--- a/robot/control/Inc/MicroPackets.hpp
+++ b/robot/control/Inc/MicroPackets.hpp
@@ -141,7 +141,7 @@ struct KickerCommand {
     ShootMode shootMode;     /**< 1-bit encoding of whether the kicker should perform a ground kicker or a chip */
     TriggerMode triggerMode; /**< 2-bit encoding of when kicker should kick (none, immediately, on breakbeam, cancel) */
 
-    uint8_t kickStrength;    /**< 4-bit encoding of the kick's strength (0: min strength - 15: max strength) */
+    uint8_t kickStrength;    /**< 8-bit encoding of the kick's strength (0-255) */
 };
 
 /** @struct KickerInfo
--- a/robot/control/Inc/MicroPackets.hpp
+++ b/robot/control/Inc/MicroPackets.hpp
@@ -10,128 +10,168 @@
 // The "consumer" of the data must check if it's valid and
 // the data isn't too old (if applicable)
 
-// Robot body velocities and dribbler commands
-// Radio -> Motion Control
+/** @struct MotionCommand
+ * Contains robot body velocities and dribbler commands
+ *
+ * Passed from @ref RadioModule to @ref MotionControlModule
+ */
 struct MotionCommand {
-    bool isValid = false;
-    uint32_t lastUpdate;
+    bool isValid = false; /**< Stores whether given data is valid  */
+    uint32_t lastUpdate;  /**< Time that MotionCommand was sent (milliseconds) */
 
-    float bodyXVel; // m/s
-    float bodyYVel; // m/s
-    float bodyWVel; // rad/s
+    float bodyXVel;       /**< x-component of robot's linear velocity (m/s) */
+    float bodyYVel;       /**< y-component of robot's linear velocity (m/s) */
+    float bodyWVel;       /**< Angular velocity of robot about z axis (rad/s) */
 
-    // Limited from 0-128
-    uint8_t dribbler; // lsb
+    uint8_t dribbler;     /**< 7-bit encoding of dribble rotation (0: no spin - 128: max spin)
+                          @note **This is limited from 0-128** */
 };
 
-// Duty cycle commands for individual motors
-// MotionControl -> FPGA
+/** @struct MotorCommand
+ * Contains duty cycle commands for individual motors
+ *
+ * Passed from @ref MotionControlModule to @ref FPGAModule
+ */
 struct MotorCommand {
-    bool isValid = false;
-    uint32_t lastUpdate;
+    bool isValid = false; /**< Stores whether given data is valid  */
+    uint32_t lastUpdate;  /**< Time at which MotorCommand was last updated (milliseconds) */
 
-    float wheels[4]; // % max duty cycle (-1 to 1)
-    uint16_t dribbler; // lsb (0 to 128)
+    float wheels[4];      /**< % max duty cycle for each wheel motor (-1 to 1) */
+    uint16_t dribbler;    /**< 7-bit encoding of dribble rotation (0: no spin - 128: max spin)
+                          @note **This is limited from 0-128** */
 };
 
-// Current and enc data
-// FPGA -> MotionControl
+/** @struct MotorFeedback
+ * Contains current and encoder data for each motor
+ *
+ * Passed from @ref FPGAModule to @ref MotionControlModule
+ */
 struct MotorFeedback {
-    bool isValid = false;
-    uint32_t lastUpdate;
+    bool isValid = false; /**< Stores whether given data is valid  */
+    uint32_t lastUpdate;  /**< Time at which MotorFeedback was last updated (milliseconds) */
 
-    float encoders[4]; // rad / s
-    float currents[4]; // amp
+    float encoders[4];    /**< Encoder readings from each wheel motor (rad/s)  */
+    float currents[4];    /**< Current readings from each wheel motor (amps)  */
 };
 
-// Accelerations and angular velocities
-// IMU -> MotionControl
+/** @struct IMUData
+ * Contains linear acceleration and angular velocity readings from IMU
+ *
+ * Passed from @ref IMUModule to @ref MotionControlModule
+ */
 struct IMUData {
-    bool isValid = false;
-    bool initialized = false;
-    uint32_t lastUpdate;
+    bool isValid = false;     /**< Stores whether given data is valid  */
+    bool initialized = false; /**< Stores whether IMU has been initialized */
+    uint32_t lastUpdate;      /**< Time at which IMUData was last updated (milliseconds) */
 
-    float accelerations[3]; // X Y Z (m/s2)
-    float omegas[3]; // X Y Z (rad/s)
+    float accelerations[3];   /**< Linear acceleration on axes [X,Y,Z] (m/s^2) */
+    float omegas[3];          /**< Angular velocities on axes [X,Y,Z] (rad/s) */
 };
 
-// Battery voltage and whether it's at a critical level
-// Battery -> MotionControl
-// Battery -> LED
-// Battery -> Radio
+/** @struct BatteryVoltage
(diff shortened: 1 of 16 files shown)


robot/control/Inc/motion-control/RobotEstimator.hpp
```diff
@@ -43,21 +68,68 @@ class RobotEstimator {
     void getState(Eigen::Matrix<float, numStates, 1>& state);
 
 private:
-    static constexpr float processNoise = 0.05;
-    static constexpr float encoderNoise = 0.04;
-    static constexpr float gyroNoise = 0.005;
-    static constexpr float initCovariance = 0.1;
 
+    static constexpr float processNoise = 0.05;  /**< Random noise caused by external influences on the robot */
```
JNeiger, line 72: I would say processNoise represents any unmodeled influence on the states. It could be assumptions about the constant velocity built into the transition matrices, modeling errors on the actual robot properties like weight, it could be completely unmodeled things like wheel slip.

robot/lib/Inc/drivers/Battery.hpp
```diff
@@ -1,28 +1,47 @@
 #pragma once
 
 #include "cstdint"
-
+/** @class Battery
```
JNeiger, line 4: nit: Add a newline before the comment

robot/lib/Inc/drivers/Battery.hpp
```diff
@@ -1,28 +1,47 @@
...
+ * Interfaces with analog pin to get voltage-related data
+ *
+ * @note As of May 2020, this class does not currently work, as it does not read analog pin
```
JNeiger, line 7: Specially, we do not have an analog in driver

robot/lib/Inc/drivers/Battery.hpp
```diff
@@ -1,28 +1,47 @@
...
 
 private:
-    // 5-cell lipo * 4.3 v per cell = 21.5 battery volts
-    // 68k and 10k ohm voltage divider, analog in read votlage 2.756
+    /**
+     * The maximum voltage the battery can safely hold
+     *
+     * Calculations:
+     *  - 5-cell lipo * 4.3 v per cell = 21.5 battery volts
+     *  - 68k and 10k ohm voltage divider, analog in read voltage 2.756
```
JNeiger, line 27: This has been here a long time, we should double check and make sure these are the resistors still

robot/lib/Inc/drivers/Battery.hpp
```diff
@@ -1,28 +1,47 @@
...
+    void update();             /**< Updates reading from analog pin*/
 
-    float getBattPercentage();
+    float getBattPercentage(); /**< Returns the last battery percentage read */
 
-    uint8_t getRaw();
+    uint8_t getRaw();          /**< Returns battery voltage as a raw 8-bit integer */
 
-    bool isBattCritical();
+    bool isBattCritical();     /**< Stores whether battery is critical (at or below 0%) */
```
JNeiger, line 19: Is this the same as `@return`?

robot/lib/Inc/drivers/KickerBoard.hpp
```diff
@@ -132,17 +132,53 @@ class KickerBoard : public AVR910 {
...
+     */
     const uint8_t isChargedCutoff = 230;
 
+    /**
+     * Stores whether the breakbeam has been tripped
+     */
     bool _ball_sensed = false;
+
+    /**
+     * Stores whether Kicker is responding back with voltage values
```
JNeiger, line 146: I would be more specific and say
```suggestion
     * Stores whether Kicker is responding back with non-zero voltage values
```

robot/lib/Inc/drivers/KickerBoard.hpp
```diff
@@ -132,17 +132,53 @@ class KickerBoard : public AVR910 {
...
+     */
     bool _is_healthy = true;
 
+    /**
+     * Current voltage stored in Kicker, mapped onto 8 bits (0-255)
```
JNeiger, line 151: In units of Volts

robot/lib/Inc/drivers/KickerBoard.hpp
```diff
@@ -132,17 +132,53 @@ class KickerBoard : public AVR910 {
...
+     */
     bool _cancel_kick    = true;
+
+    /**
+     * Stores whether the kicker board can safely charge the capacitors
+     */
     bool _charge_allowed = false;
+
+    /**
+     * 4-bit encoding of the kick's strength (0: min strength - 15: max strength)
```
JNeiger, line 181: 0-256 here, it's converted just before send

robot/control/Inc/modules/BatteryModule.hpp
```diff
@@ -6,19 +6,44 @@
...
+     * Priority used by RTOS
+     *
+     * Uses the default priority of 1.
+     */
     static constexpr int kPriority = 1;
 
+    /**
+     * Constructor for BatteryModule
+     *
+     * @param batteryVoltage Packet of data containing data on battery voltage and critical status
```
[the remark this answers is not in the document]
JNeiger: This is applicable to all module comments

robot/control/Inc/modules/BatteryModule.hpp
```diff
@@ -6,19 +6,44 @@
...
+     */
     explicit BatteryModule(LockedStruct<BatteryVoltage>& batteryVoltage);
 
+    /**
+     * Code to run when called by RTOS
```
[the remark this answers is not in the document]
JNeiger: This is applicable to all module comments

robot/control/Inc/modules/LEDModule.hpp
```diff
@@ -11,50 +11,89 @@
...
+     *
+     * Checks for Radio, FPGA, and Kicker errors, and activates dotStar LEDs accordingly
+     */
     void entry() override;
 
-    // Specific LED pattern for fpga initialization
+    /**
+     * Set specific LED pattern for fpga initialization
+     */
     void fpgaInitialized();
```
JNeiger, line 68: For these `*Initialized()` functions, fpga turns on LED1, radio turns on LED2, kicker turns on LED3, fully turns on LED4, resulting in a loading bar if you watch the startup sequence

robot/control/Inc/modules/IMUModule.hpp
```diff
@@ -7,17 +7,43 @@
...
 
+/**
+ * Module interfacing with IMU and handling IMU status
+ */
 class IMUModule : public GenericModule {
 public:
-    // How many times per second this module should run
-    static constexpr float kFrequency = 200.0f; // Hz
+    /**
+     * Number of times per second (frequency) that BatteryModule should run (Hz)
```
Arvind-Srinivasan, line 16: Shouldn't this be IMUModule?

robot/lib/Inc/drivers/Battery.hpp
```diff
@@ -1,28 +1,47 @@
...
-    // 5-cell lipo * 3.0 v per cell = 15.0 battery volts
-    // 68k and 10k ohm voltage divider, analog in read votlage 1.923
+    /**
+    * The minimum voltage the battery can safely hold
+    *
+    * Calculations:
+    *  - 5-cell lipo * 4.3 v per cell = 21.5 battery volts
+    *  - 68k and 10k ohm voltage divider, analog in read voltage 1.923
+    */
     const float MIN_SAFE_BATT_VOLTAGE_READ = 1.923;
```
Arvind-Srinivasan, line 40: Is this not dependent on our current battery, which is changing?

robot/control/Inc/MicroPackets.hpp
```diff
@@ -10,128 +10,167 @@
...
-    uint32_t lastUpdate;
+    bool isValid = false; /**< Stores whether given data is valid  */
+    uint32_t lastUpdate;  /**< Time that MotionCommand was sent (milliseconds) */
 
-    float bodyXVel; // m/s
-    float bodyYVel; // m/s
-    float bodyWVel; // rad/s
+    float bodyXVel;       /**< x-component of robot's linear velocity (m/s) */
+    float bodyYVel;       /**< y-component of robot's linear velocity (m/s) */
+    float bodyWVel;       /**< Angular velocity of robot about z axis (rad/s) */
```
Arvind-Srinivasan, line 24: Maybe say for reference "Angular velocity (w-component)"

robot/lib/Inc/drivers/Battery.hpp
Arvind-Srinivasan: Those resistors are connect as of control 3.4

robot/control/Inc/MicroPackets.hpp
JNeiger: The W was supposed to represent omega here when I originally wrote it. About the Z axis as commented makes sense or the the r component of the pqr vector for angular velocity makes sense. I'm not sure I agree with the w component though

robot/control/Inc/modules/IMUModule.hpp
tcontis: Yes! I went through modules again and found a few errors which I'll add in a fix commit.

robot/control/Inc/MicroPackets.hpp
tcontis: I agree with @JNeiger 's assessment here. W (omega, really) is a variable representing angular velocity. Truly, it should be a vector with specific components (theta, phi, or whatever greek symbols you most prefer)

robot/lib/Inc/drivers/Battery.hpp
tcontis: I accidentally treated this as a variable, but it is a function. I'll put returns instead (rather than stores)

robot/control/Inc/modules/BatteryModule.hpp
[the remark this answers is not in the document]
tcontis: I'm debating using the copy docs and inherit functionality, because my rationale is that it would be easier to just see the comments right in front of your eyes when you are looking through a file. Although, I suppose you might make the case that one can just take a look at the parent class

robot/control/Inc/modules/RadioModule.hpp
```diff
@@ -35,6 +67,9 @@ class RadioModule : public GenericModule {
     LockedStruct<MotionCommand>& motionCommand;
     LockedStruct<RadioError>& radioError;
 
+    /**
+     * Object which sends and receives radio packets
+     */
```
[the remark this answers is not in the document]
tcontis: How about "General radio driver interface acting as a middle man to send and receive radio packets" ?

robot/control/Inc/motion-control/DribblerController.hpp
```diff
@@ -20,13 +25,21 @@ class DribblerController {
 private:
```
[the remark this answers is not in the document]
tcontis: Is LSB equivalent to 1 in most of these contexts? Could I then but (0-255) like for all the others?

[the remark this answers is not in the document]
JNeiger: It's between 0 and 128 for the dribber. In most things it's LSB/[other unit]. So LSB/(m/s^2) for Accels to convert the output between their made up unit scale and m/s^2. Here, there's some relationship between LSB and duty cycles, but it doesn't really matter at all. In those cases, I've seen people say it's in LSB to signify that it's not a metric unit, but it acts like a unit.

LSB is least significant bit. For the accel example, for every 1 increment in the LSB number, the actual acceleration it measures would increase by 1/1024 m/s^2 (or whatever that conversion is)

robot/control/Inc/modules/RadioModule.hpp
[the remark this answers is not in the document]
JNeiger: Yeah, that sounds pretty solid

robot/control/Inc/modules/BatteryModule.hpp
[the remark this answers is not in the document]
JNeiger: Yeahh, it's could go either way. Honestly, pick a version and stick with it and everyone else will follow suite. I just wanted to make sure you knew about those options

robot/lib/Inc/drivers/Battery.hpp
Arvind-Srinivasan: @mmaisonneuve3 can you confirm the batteries we are ordering and what their voltages are?

PR #112, opened 2020-07-09
LEDModule update

tcontis:
Added new infrastructure based on adding Error structs of led colors to a cycling queue for the dotStar LEDs. Displaying LED colors while the LEDModule is running is now handled with the queue and the displayErrors() method.
The LEDModule will flash an error on and off for specified counts before moving to the new one

### DIFF
--- a/robot/control/Inc/MicroPackets.hpp
+++ b/robot/control/Inc/MicroPackets.hpp
@@ -110,7 +110,6 @@ struct RadioError {
     bool initialized = false; /**< Stores whether Radio has been initialized */
     uint32_t lastUpdate;      /**< Time at which RadioError was last updated (milliseconds) */
 
-    bool hasError;                 /**< Stores whether Radio has an error */
     bool hasConnectionError;       /**< Stores if Radio is having trouble connecting to WiFi */
     bool hasSoccerConnectionError; /**< Stores if Radio is having trouble communicating with Soccer */
 };
--- a/robot/control/Inc/MicroPackets.hpp
+++ b/robot/control/Inc/MicroPackets.hpp
@@ -110,8 +110,9 @@ struct RadioError {
     bool initialized = false; /**< Stores whether Radio has been initialized */
     uint32_t lastUpdate;      /**< Time at which RadioError was last updated (milliseconds) */
 
-    bool hasError;            /**< Stores whether Radio has an error */
-    bool hasConnectionError;   /**< Stores if Radio is having trouble connecting to WiFi */
+    bool hasError;                 /**< Stores whether Radio has an error */
+    bool hasConnectionError;       /**< Stores if Radio is having trouble connecting to WiFi */
+    bool hasSoccerConnectionError; /**< Stores if Radio is having trouble communicating with Soccer */
 };
 
 /** @struct RobotID
--- a/robot/control/Inc/MicroPackets.hpp
+++ b/robot/control/Inc/MicroPackets.hpp
@@ -111,6 +111,7 @@ struct RadioError {
     uint32_t lastUpdate;      /**< Time at which RadioError was last updated (milliseconds) */
 
     bool hasError;            /**< Stores whether Radio has an error */
+    bool hasConnectionError;   /**< Stores if Radio is having trouble connecting to WiFi */
 };
 
 /** @struct RobotID
--- a/robot/control/Inc/modules/LEDModule.hpp
+++ b/robot/control/Inc/modules/LEDModule.hpp
@@ -42,7 +42,9 @@ enum errorColors_1 : uint32_t {
  */
 enum errorColors_2 : uint32_t {
     RADIO_BOOT_FAIL = 0x0000FF, // RED
-    RADIO_CONN_WIFI_FAIL = 0x0080FF  // ORANGE
+    RADIO_CONN_WIFI_FAIL = 0x0080FF,  // ORANGE
+    KICKER_BOOT_FAIL = 0x000FF, // RED
+    FPGA_BOOT_FAIL = 0x000FF // RED
 };
 
 /**
@@ -202,7 +204,12 @@ private:
     const struct Error ERR_RADIO_CONN_FAIL = {errorColors_0::RADIO_ERROR,
                                               errorColors_1::FATAL,
                                               errorColors_2::RADIO_CONN_WIFI_FAIL};
-
+    const struct Error ERR_KICKER_BOOT_FAIL = {errorColors_0::KICKER_ERROR,
+                                               errorColors_1::FATAL,
+                                               errorColors_2::KICKER_BOOT_FAIL};
+    const struct Error ERR_FPGA_BOOT_FAIL = {errorColors_0::FPGA_ERROR,
+                                             errorColors_1::FATAL,
+                                             errorColors_2::FPGA_BOOT_FAIL};
 
     /**
      * Vector of colors for dotStars to display sequentially based on current errors
--- a/robot/control/Inc/modules/LEDModule.hpp
+++ b/robot/control/Inc/modules/LEDModule.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <algorithm>
 #include <array>
 #include <memory>
 #include <vector>
@@ -40,14 +41,14 @@ enum errorColors_1 : uint32_t {
  * Third LED in error display format: Category / Level / Info
  */
 enum errorColors_2 : uint32_t {
-    RADIO_BOOT_FAIL = 0x0000FF // RED
+    RADIO_BOOT_FAIL = 0x0000FF, // RED
+    RADIO_CONN_FAIL = 0x0080FF // ORANGE
 };
 
 /**
  * Struct to store error names and LED values in colorQueue
  */
 struct Error {
-    const std::string name;
     const uint32_t led0;
     const uint32_t led1;
     const uint32_t led2;
@@ -176,12 +177,12 @@ private:
     /**
      * Add error to colorQueue
      */
-    void addError(Error e);
+    void addError(Error newError);
 
     /**
      * Remove error from colorQueue
      */
-    void removeError(std::string name);
+    void removeError(Error error);
 
     const static uint16_t IOExpanderErrorLEDMask = 0xFF00;
 
@@ -194,6 +195,10 @@ private:
     LockedStruct<RadioError>& radioError;
 
     DigitalOut dotStarNCS;
+
+    const struct Error ERR_RADIO_BOOT_FAIL = {errorColors_0::RADIO_ERROR, errorColors_1::FATAL, errorColors_2::RADIO_BOOT_FAIL};
+    const struct Error ERR_RADIO_CONN_FAIL = {errorColors_0::RADIO_ERROR, errorColors_1::FATAL, errorColors_2::RADIO_CONN_FAIL};
+
 
     /**
      * Vector of colors for dotStars to display sequentially based on current errors
--- a/robot/control/Inc/modules/LEDModule.hpp
+++ b/robot/control/Inc/modules/LEDModule.hpp
@@ -233,7 +233,7 @@ private:
     std::vector<Error> colorQueue;
 
     /**
-     * Current index of displayed color
+     * Current index of colorQueue
      */
     size_t index = 0;
 
--- a/robot/control/Inc/modules/LEDModule.hpp
+++ b/robot/control/Inc/modules/LEDModule.hpp
@@ -18,10 +18,10 @@
  * First LED in error display format: Category / Level / Info
  */
 enum errorColors_0 : uint32_t {
-    RADIO_ERROR = 0x0080FF;  // ORANGE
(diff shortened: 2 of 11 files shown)


robot/control/Inc/modules/LEDModule.hpp
```diff
@@ -11,6 +13,51 @@
 #include "MicroPackets.hpp"
 #include "drivers/MCP23017.hpp"
 
+
+/**
+ * Constants for DotStar LEDs (BGR)
+ * First LED in error display format: Category / Level / Info
+ */
+enum errorColors_0 : uint32_t {
```
JNeiger, line 21: I would call this enum `DeviceErrorColors` or something since you're treating the three led's as a set.

robot/control/Inc/modules/LEDModule.hpp
```diff
@@ -11,6 +13,51 @@
...
+    FPGA_ERROR = 0x00FFFF,    // YELLOW
+    KICKER_ERROR = 0x00FF00, // GREEN
+    IMU_ERROR = 0xFF0000   // BLUE
+};
+
+/**
+ * Constants for DotStar LEDs (BGR)
+ * Second LED in error display format: Category / Level / Info
+ */
+enum errorColors_1 : uint32_t {
```
JNeiger, line 32: `ErrorLevelColors` or something?

robot/control/Inc/modules/LEDModule.hpp
```diff
@@ -11,6 +13,51 @@
...
+    ERR = 0x0080FF, // ORANGE
+    WARN = 0xFFFF00,  // YELLOW
+    INFO = 0xFF0000  // BLUE
+};
+
+/**
+ * Constants for DotStar LEDs (BGR)
+ * Third LED in error display format: Category / Level / Info
+ */
+enum errorColors_2 : uint32_t {
```
JNeiger, line 43: `ErrorTypeColors` or something?

robot/control/Src/modules/LEDModule.cpp
```diff
@@ -21,8 +23,8 @@ LEDModule::LEDModule(LockedStruct<MCP23017>& ioExpander,
 void LEDModule::start() {
     auto ioExpanderLock = ioExpander.lock();
     ioExpanderLock->init();
-    setColor(0xFFFFFF, 0xFFFFFF);
-
+    setColor(0xFFFFFF, 0xFFFFFF, 0xFFFFFF);
+    index = 0;
```
JNeiger, line 27: I think this makes sense more in the constructor instead of here

robot/control/Src/modules/LEDModule.cpp
```diff
@@ -168,4 +198,52 @@ void LEDModule::setColor(uint32_t led0, uint32_t led1) {
...
+        setColor(0x000000,0x000000, 0x000000);
+    } else if (!lightsOn && framesOffCounter >= framesOff) {
+        // If error lights have been off long enough, switch them back on to next error color
+        framesOffCounter = 0;
+        lightsOn = true;
+
+        // Make sure index loops back if it has reached end of vector
+        if (++index > colorQueue.size()-1) {
+            index = 0;
+        }
```
JNeiger, line 222: This took me a second to figure out, I usually see it as `index = (index + 1) % colorQueue.size();`

robot/control/Src/modules/LEDModule.cpp
```diff
@@ -168,4 +198,52 @@ void LEDModule::setColor(uint32_t led0, uint32_t led1) {
...
+    colorQueue.push_back(newError);
+}
+
+void LEDModule::removeError(Error target) {
+    colorQueue.erase(std::remove_if(colorQueue.begin(), colorQueue.end(),
+            [&](Error err) {
+                return err.led0 == target.led0 &&
+                       err.led1 == target.led1 &&
+                       err.led2 == target.led2;
+           }), colorQueue.end());
```
JNeiger, line 248: This makes me nervous in an embedded environment since I haven't tried any of these algorithms/lambdas in this context. Since this is getting checked every frame, it may cause some problems. It's definitely something to check

robot/control/Inc/MicroPackets.hpp
```diff
@@ -110,7 +110,9 @@ struct RadioError {
     bool initialized = false; /**< Stores whether Radio has been initialized */
     uint32_t lastUpdate;      /**< Time at which RadioError was last updated (milliseconds) */
 
-    bool hasError;            /**< Stores whether Radio has an error */
+    bool hasError;                 /**< Stores whether Radio has an error */
```
JNeiger, line 113: Can you change the comment to specify it's a logical or of the hasConnectionError/hasSoccerConnectionError? Either that or completely remove this since the other errors are more specific

robot/control/Src/radio/RadioLink.cpp
```diff
@@ -83,5 +85,7 @@ bool RadioLink::receive(KickerCommand& kickerCommand,
     motionCommand.bodyWVel = static_cast<float>(control->bodyW) / rtp::ControlMessage::VELOCITY_SCALE_FACTOR;
     motionCommand.dribbler = control->dribbler;
 
+    radioConnected = true;
```
JNeiger, line 88: So connected is based off receiving a packet?

robot/lib/Inc/drivers/ISM43340.hpp
```diff
@@ -371,7 +371,8 @@ class ISM43340 : public GenericRadio {
 
     void pingRouter();
 
-    bool isConnected() const { return isInit; }
+    bool isInitialized() const { return isInit; }
+    bool isConnected() { return connected; }
```
JNeiger, line 375: `isConnected` and maybe `isInitialized` are both generic enough that we can put this into the `GenericRadio` interface. This way `isConnected` in `RadioLink` is not based on receiving a packet, but on connecting to the wifi network

robot/control/Inc/modules/LEDModule.hpp
```diff
@@ -130,9 +194,62 @@ class LEDModule : public GenericModule {
...
+    // IMU
+    const struct Error ERR_IMU_BOOT_FAIL =  {errorColors_0::IMU_ERROR,
+                                             errorColors_1::FATAL,
+                                             errorColors_2::BOOT_FAIL};
+
+
+    /**
+     * Vector of colors for dotStars to display sequentially based on current errors
+     */
+    std::vector<Error> colorQueue;
```
JNeiger, line 229: I wonder if there's a way to make this a static list with booleans/mask to turn on and off errors. This way you don't have to copy and remove from the vector continuously since each ERR should only exist once

robot/control/Src/modules/LEDModule.cpp
```diff
@@ -63,53 +68,79 @@ void LEDModule::entry() {
...
 }
 
 void LEDModule::fpgaInitialized() {
     // neo pixel stuff
-    setColor(0xFFFF00, 0xFFFFFF);
     leds[0] = 1;
 }
 
 void LEDModule::radioInitialized() {
     // neo pixel stuff
```
JNeiger, line 133: Can you remove the `//neo pixel stuff` comment as well? for all these

robot/control/Src/modules/LEDModule.cpp
tcontis: That's actually a very concise way to write it, since modular math always wraps things back around. I like it!

robot/control/Inc/modules/LEDModule.hpp
tcontis: I think that what you gain in memory efficiency you lose in having to find the next element in the array which is true

robot/control/Src/modules/LEDModule.cpp
tcontis: I don't really see too much harm in this lambda because it's just simply checking whether given sets of constants are equals from different error structs. However, I'll make note to test this for safety

robot/control/Src/radio/RadioLink.cpp
tcontis: Yes. Connected over WiFi is based on having received a packet. That's probably not a good measure, though. It's probably redundant with the soccer connection error which is based on there being a steady stream of packets sent back and forth.

Idk how to detect whether the radio is simply just connected to wifi or whether this is even useful at all. Since soccer connection error could either mean radio is not connected to wifi or that soccer timed out, I kind of wanted do distinguish the two. Maybe it's not necessary

robot/lib/Inc/drivers/ISM43340.hpp
tcontis: I don't quite get the second sentence. How would moving these methods to the interface help? Wouldn't they still have to defined individually in the subclasses?

JNeiger: If you move it up to the interface, we can call it from the radio link class and convert that `radioConnected` flag to be representing the `isConnected()` from the driver (in our case that is connected to the network) instead of just representing receiving a packet. Same with `isInitialized()`. It basically allows errors from the drivers to get pushed up into the higher level things, anything not in the interface is locked into the driver and can never be exposed

robot/control/Src/modules/LEDModule.cpp
```diff
@@ -168,4 +195,44 @@ void LEDModule::setColor(uint32_t led0, uint32_t led1) {
...
+        // If error lights have been off long enough, switch them back on to next error color
+        framesOffCounter = 0;
+        lightsOn = true;
+
+        // Make sure index loops back if it has reached end of vector
+        // Since we know there is at least one error, we can safely use a while loop to find it.
+        index = (index + 1) % errToggles.size();
+        while(!errToggles.at(index)) {
+            index = (index + 1) % errToggles.size();
+        }
```
JNeiger, line 221: Super nit: do while works here to find the next error in the list

robot/control/Inc/MicroPackets.hpp
tcontis: `hasError` is just meant to indicate whether the radio has any error at all. It's basically the or of all other radio errors.

robot/control/Src/radio/RadioLink.cpp
```diff
@@ -83,5 +86,7 @@ bool RadioLink::receive(KickerCommand& kickerCommand,
     motionCommand.bodyWVel = static_cast<float>(control->bodyW) / rtp::ControlMessage::VELOCITY_SCALE_FACTOR;
     motionCommand.dribbler = control->dribbler;
 
+    radioConnected = radio->isConnected();
```
JNeiger, line 89: I'm trying to think if it matters if this is at the top of this function or not. Specifically, `radio->isAvailable` and `radio->receive` `if` statements would both return false, `radio->isConnected` is false so it will never reach this line to update `radioConnected`. 

In our current implementation of the `ISM43340`, we only update `connected` once, so it doesn't make a difference, but if future changes resulted in the ISM driver (or any other radio driver) being able to update `connected` on the fly. Loosing connection to the wifi should result in the `connected` variable changing, which will result in `radioConnected` here never updating because it never reaches this line since it is reasonable that `radio->isAvailable` and `radio->receive` both fail their `if` statements.

TLDR: We should move this line to the top of the function above any returns.

JNeiger: Thinking about this more, just remove the radioConnected variable and return radio->isConnected in that getter function. That solves all the problems and is the same behavior

tcontis: That's a good way to do it

robot/control/Inc/modules/LEDModule.hpp
```diff
@@ -130,9 +189,68 @@ class LEDModule : public GenericModule {
...
+     * Number of frames (1/kPeriod seconds) for which the error lights will be on, to cycle through error LEDs
+     */
+    int framesOn = 2;
+    int framesOnCounter = 0;
+
+    /**
+     * Number of frames (1/kPeriod seconds) for which the error lights will be off, to cycle through error LEDs
+     */
+    int framesOff = 3;
+    int framesOffCounter = 0;
```
Arvind-Srinivasan, line 247: Is this enough time to be visible? Its on for about 200 ms and off for 300 ms?

robot/control/Src/modules/LEDModule.cpp
```diff
@@ -168,4 +195,44 @@ void LEDModule::setColor(uint32_t led0, uint32_t led1) {
     dotStarNCS.write(false);
     dotStarSPI.lock()->transmit(data);
     dotStarNCS.write(true);
-}
\ No newline at end of file
+}
+
+void LEDModule::displayErrors() {
+    // If there are no errors, then there is no point in displaying anything
+    if (std::find(errToggles.begin(), errToggles.end(), true) == errToggles.end()) {
```
Arvind-Srinivasan, line 202: Could you use [none_of](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of) here?

tcontis: Yes, that's a nice expression!

Issue #114, opened 2020-07-18
Swap DotStar from BGR to RGB in code


JNeiger:
Right now the int32 holding the RGB value for the dot star is XXBBGGRR, which is difficult to use. It should be XXRRGGBB. The code that interacts with the dot star needs to change and the constants for the color need to change as well

PR #115, opened 2020-07-19
BGR to RGB

tcontis:
Fixed colors to be RGB by default, added conversion function to BGR.
Addresses Issue #114

### DIFF
--- a/robot/control/Src/modules/LEDModule.cpp
+++ b/robot/control/Src/modules/LEDModule.cpp
@@ -55,8 +55,6 @@ void LEDModule::entry() {
             for (int i = 0; i < 5; i++) {
                 errors |= (1 << motors[i]);
             }
-
-            setColor(0x0000FF, 0xFFFFFF, 0xFFFFFF);
         } else {
             for (int i = 0; i < 5; i++) {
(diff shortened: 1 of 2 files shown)


robot/control/Src/modules/LEDModule.cpp
```diff
@@ -167,6 +165,10 @@ void LEDModule::setColor(uint32_t led0, uint32_t led1, uint32_t led2) {
 
     std::vector<uint8_t> data;
 
+    led0 = RGB_TO_BGR(led0);
+    led1 = RGB_TO_BGR(led1);
+    led2 = RGB_TO_BGR(led2);
```
JNeiger, line 170: I would just change the shifting done on line 178-190. I'm not sure it adds much doing the flip

tcontis: Probably not. I was thinking of doing this, but now come to think of it, the improvement in readability sacrifices efficiency, 6 bitwise ANDs, 5 bitwise shifts, and 3 intermediate variables vs only 3 bitwise ANDs and 3 bitwise shifts.

PR #120, opened 2020-10-05
Kicker integration

Arvind-Srinivasan:
Fixed kicker firmware and adjusted build process


kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
-    // such that it doesn't change halfway through the kick
-    current_kick_type_is_kick = is_kick;
+  // Wait for ADSC bit to clear
+  while (ADCSRA & _BV(ADSC))
+    ;
 
-    // start timer to enable the kick FSM processing interrupt
-    TCCR0B |= _BV(CS01); // No prescale
+  // ADHC will range from 0 to 255 corresponding to 0 through VCC
+  return (ADCH << 1);
```
JNeiger, line 138: What's this left shift from?

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
-    // so we should swap
-    if (ball_sense_change_count > BALL_SENSE_MAX_SAMPLES) {
-        ball_sense_change_count = 0;
-
-        ball_sensed = !ball_sensed;
+    // don't run the adc every loop
+    if (time % 1000 == 0) {
+      int voltage_accum =
+        (255 - kalpha) * last_voltage_ + kalpha * (get_voltage());
+      last_voltage_ = voltage_accum / 255;
```
JNeiger, line 192: Is this actually noisy?

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
+          break;
+        case 4:
+          PORTA &= 0b11000011;
+          break;
+        case 5:
+          PORTA &= 0b11000001;
+          break;
+        default:
+          PORTA &= 0xFF;
+      }
```
JNeiger, line 217: +1 on switch, much much cleaner than bit magic

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
     }
-}
+    time++;
```
JNeiger, line 219: I'll call this out, but it really doesn't matter, every time `time` overflows, it won't be exactly `1000` frames apart. I would just leave it though

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
-        !charge_allowed ||
-        !command.commanded_charge) {
 
-        HAL_ClearPin(LT_CHARGE);
+    if (!(PIND & _BV(LT_DONE_N)) || last_voltage_ > 239 || !charge_allowed_ ||
+        !charge_commanded_) {
+      PORTD &= ~(_BV(LT_CHARGE));
+    } else if (last_voltage_ < 232 && charge_allowed_ && charge_commanded_) {
+      PORTD |= _BV(LT_CHARGE);
+    }
```
JNeiger, line 230: There's a lot of terms here and it's very important this is correct, could this get commented?

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
 
-    // Charge if we are too low
-    } else if (current_voltage < 239 &&
-               charge_allowed &&
-               command.commanded_charge) {
+    if (PINB & _BV(N_KICK_CS_PIN)) {
+      byte_cnt = 0;
+    }
```
JNeiger, line 234: `byte_cnt` isn't being used anywhere

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
 
-        HAL_SetPin(LT_CHARGE);
+    bool bs = PINA & _BV(BALL_SENSE_RX);
+    if (ball_sensed_) {
```
JNeiger, line 237: This can be condensed using `ball_sensed_ ^ bs` then `ball_sense_change_count_++` else `ball_sense_change_count_ = 0`. See old code

kicker/main.c
```diff
@@ -316,195 +417,98 @@ ISR(SPI_STC_vect) {
...
+    charge_commanded_ = true;
+  } else {
+    charge_commanded_ = false;
+  }
+
+  if (kick_activation == KICK_ON_BREAKBEAM) {
+    kick_on_breakbeam_ = true;
+    kick_on_breakbeam_strength_ = kick_power;
+  } else if (kick_activation == KICK_IMMEDIATE) {
+    kick(kick_power,false);
```
JNeiger, line 507: You call kick in the interrupt. You'll have to measure it, but at least for the attiny, the time it takes to run this was actually an order of magnitude longer than the SPI transaction just to set global bools and a few ints. If that's still the case, I wouldn't call this and be doing the floating point calculations we do in `kick`. There was a `command.kick_immediate` variable and that kick check in main to avoid this.

kicker/main.c
```diff
@@ -316,195 +417,98 @@ ISR(SPI_STC_vect) {
...
+ * short!
+ */
+uint8_t execute_cmd(uint8_t cmd) {
+  // if we don't change ret_val by setting it to voltage or
+  // something, then we'll just return the command we got as
+  // an acknowledgement.
+  bool allow_charge = !!(cmd & (1 << 4));
+  uint8_t kick_power = (cmd & 0xF) << 4;
+  uint8_t kick_activation = cmd & (3 << 5);
+  bool use_chip = !!(cmd & (1 << 7));
```
JNeiger, line 495: Have the chip stuff implemented, but commented out. It'll make it easier later to come back and put it back in

kicker/main.c
```diff
@@ -1,302 +1,403 @@
...
+volatile int32_t timer_cnts_left_ = -1;
+volatile int32_t post_kick_cooldown_ = -1;
+volatile int32_t kick_wait_ = -1;
 
-// Corresponds to the values of the kick_type_is_kick
-#define IS_KICK true
-#define IS_CHIP false
+//Check for chip type
+volatile bool kick_type_is_chip_ = false;
+volatile bool kick_type_is_chip_command_ = false;
```
JNeiger, line 54: I don't see these used, but the purpose is so that during a kick/chip, if the kick type is changed, we don't swap which FET we are opening or closing.

AKA
```
command immediate kick
open kick fet
command immediate chip
close chip fet
```
If `kick_type_is_chip` is locked on calls to `kick()`, and all commands are saved into `kick_type_is_chip_command` then this problem is eliminated.

kicker/main.c
JNeiger: I see you guys disable SPI in init, can resolve this thread

kicker/kicker_commands.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
JNeiger, line 1: Assuming this is just copy-paste from before

robot/lib/Src/drivers/AVR910.cpp
```diff
@@ -334,28 +341,29 @@ int AVR910::readPartNumber() {
 
 void AVR910::chipErase() {
     auto spi_lock = lock_spi();
-
+DWT_Delay(10);
```
JNeiger, line 344: tabbing

robot/lib/Src/drivers/AVR910.cpp
```diff
@@ -367,39 +375,45 @@ void AVR910::loadMemoryPage(int highLow, char address, char data) {
...
     spi_lock->transmit(address & 0x3F); // xxxb bbbb
     spi_lock->transmit(data); // iiii iiii
     nCs_->write(1);
+    DWT_Delay(10);
 }
 
 void AVR910::writeFlashMemoryByte(int highLow, int address, char data) {
     auto spi_lock = lock_spi();
 
+DWT_Delay(10);
```
JNeiger, line 391: here too

robot/lib/Src/drivers/AVR910.cpp
```diff
@@ -367,39 +375,45 @@ void AVR910::loadMemoryPage(int highLow, char address, char data) {
...
     spi_lock->transmit(address & 0x003F);
     spi_lock->transmit(data);
     nCs_->write(1);
+    DWT_Delay(10);
 }
 
 void AVR910::writeFuseBitsLow() {
     auto spi_lock = lock_spi();
 
+DWT_Delay(10);
```
JNeiger, line 404: and here

JNeiger (commented):
I'll put this out there, kicker is one of the least touched things in the firmware stack, add in that it's the lowest level C code with the most possibility of danger, I very highly suggest adding as many comments as possible to make it extremely clear why something is done and what it represents physically. Copy-pasting all the old comments should provide a decent starting point.

Try to be extra specific will all the global variables. There's quite a few, many of which with similar names and very different functionality.

I'm biased since I wrote it, but I do like splitting everything into functions with heavy `static` use to minimize the number of global variables. I thought the very basic `HAL` was nice too, but if you guys feel it abstracted everything too much, then no worries.

kylestach (approved):

kicker/convert.py
```diff
@@ -8,7 +8,8 @@
 # This is the easy fix until a file system is built
 #
 # Example usage:
-# python3 convert.py kicker/build/bin/kicker.nib robot/lib/Inc/device-bins/kicker_bin.h KICKER_BYTES
+# python3 convert.py build/bin/kicker.nib ../robot/lib/Inc/device-bins/kicker_bin.h KICKER_BYTES
+# python3 convert.py build/bin/kicker-test.nib ../robot/lib/Inc/device-bins/kicker_bin.h KICKER_BYTES
```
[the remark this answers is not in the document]
Arvind-Srinivasan: Can you clarify why it should be in the Top level folder?

CollinAvidano:
I would really recommend people look at the organization overhaul branch. That was the structure we talked about at the end of the summer and I have a almost working branch now for mtrain since we dont quite want to move to mjackets yet it seems like. But honestly I feel like we are duplicating a lot of build systems work between all of us.

PR #121, opened 2020-10-11
LED module deadlock fix

kylestach:
The LED module was dying due to some bug in `mtrain-firmware`'s implementation of `SPI::transmit(const std::vector&)`. This fixes that, and removes some unnecessary allocation (changes `std::vector` to `std::array`).


Arvind-Srinivasan:
Looks good

PR #127, opened 2021-07-05
New radio protocol using protobuf

kylestach:
Experimental and untested but if we can support the bandwidth and memory (spoiler: we almost certainly can) this should make complex data interchange and robot configuration a lot easier.

This does allocate a big chunk of memory (16KiB right now) for the buffer; this can be trimmed down once we figure out better bounds for message sizes.

### DIFF
--- a/control/src/robocup/modules/MotionControlModule.cpp
+++ b/control/src/robocup/modules/MotionControlModule.cpp
@@ -33,22 +33,24 @@ MotionControlModule::MotionControlModule(LockedStruct<BatteryVoltage>& batteryVo
 }
 
 void MotionControlModule::entry() {
-    auto motionCommandLock = motionCommand.lock();
-    auto motorCommandLock = motorCommand.lock();
-    auto motorFeedbackLock = motorFeedback.lock();
-    auto imuDataLock = imuData.lock();
-    auto batteryVoltageLock = batteryVoltage.lock();
+    auto motion_command = motionCommand.lock().value();
+    auto motor_feedback = motorFeedback.lock().value();
+    auto imu_data = imuData.lock().value();
+    auto battery_voltage_data = batteryVoltage.lock().value();
 
-    DebugFrame frame;
-    frame.ticks = xTaskGetTickCount();
-    frame.gyro_z = imuDataLock->omegas[2];
-    frame.accel_x = imuDataLock->accelerations[0];
-    frame.accel_y = imuDataLock->accelerations[1];
+    MotorCommand motor_command;
+
+    DebugFrame frame{};
+    frame.frame_time_ms = xTaskGetTickCount();
+    frame.gyro_z_radps = imu_data.omegas[2];
+    frame.measured_acceleration[0] = imu_data.accelerations[0];
+    frame.measured_acceleration[1] = imu_data.accelerations[1];
+    frame.measured_acceleration[2] = imu_data.accelerations[2];
 
     // No radio comm in a little while. Return and die.
-    if (!motionCommandLock->isValid || !isRecentUpdate(motionCommandLock->lastUpdate)) {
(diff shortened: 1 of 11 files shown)


control/src/robocup/radio/RadioLink.cpp
```diff
@@ -14,93 +19,141 @@ void RadioLink::init() {
...
-
-        status->num_frames++;
+// Keep these variables in static memory. We'd rather get link errors if they're too big.
+// We only ever use one of these at a time
+static union {
+    rj_radio_RobotStatus robot_status_proto;
+    rj_radio_RobotCommand robot_command_proto;
+};
+// Block off a huge chunk of memory. It's really only this big because of (optional) log messages.
+static std::array<uint8_t, 16 * 1024> buffer{};
```
kylestach, line 29: Buffer memory, for visibility. The RobotStatus struct isn't small either (~5KiB), just because of the potential existence of log messages. Note that that does not mean that the wire size is 5KiB, but the in-memory version statically allocates the maximum size of 256B of data per log message * 16 log messages (intentional feature, avoids dynamic malloc).

control/src/robocup/radio/RadioLink.cpp
```diff
@@ -14,93 +19,141 @@ void RadioLink::init() {
...
-                  std::begin(to.motor_outputs));
-        std::copy(std::begin(from.encDeltas),
-                  std::end(from.encDeltas),
-                  std::begin(to.encDeltas));
-
-        status->num_frames++;
+// Keep these variables in static memory. We'd rather get link errors if they're too big.
+// We only ever use one of these at a time
+static union {
+    rj_radio_RobotStatus robot_status_proto;
```
kylestach, line 25: 5828 bytes

control/src/robocup/radio/RadioLink.cpp
```diff
@@ -14,93 +19,141 @@ void RadioLink::init() {
...
+    rj_radio_RobotCommand robot_command_proto;
```
kylestach, line 26: 69 bytes (nice)

control/src/robocup/modules/MotionControlModule.cpp
```diff
@@ -33,22 +33,24 @@ MotionControlModule::MotionControlModule(LockedStruct<BatteryVoltage>& batteryVo
...
-    auto motorFeedbackLock = motorFeedback.lock();
-    auto imuDataLock = imuData.lock();
-    auto batteryVoltageLock = batteryVoltage.lock();
-
-    DebugFrame frame;
-    frame.ticks = xTaskGetTickCount();
-    frame.gyro_z = imuDataLock->omegas[2];
-    frame.accel_x = imuDataLock->accelerations[0];
-    frame.accel_y = imuDataLock->accelerations[1];
+    auto motion_command = motionCommand.lock().value();
```
kylestach, line 36: Changes in this file do not completely fill in the new debug frame format. They were mostly intended to fix the fact that we were holding waaay too many locks throughout this whole function.

Later work will involve actually populating the new data (and providing logging facilities :smile:  )

control/src/robocup/proto/radio_protocol.proto
```diff
@@ -0,0 +1,129 @@
...
+    LEVEL_ERROR = 1;
+    LEVEL_WARNING = 2;
+    LEVEL_INFO = 3;
+    LEVEL_VERBOSE = 4;
+  }
+
+  required LogLevel level = 1;
+  required uint32 timestamp_ms = 2;
+  optional string subsystem = 3 [(nanopb).max_size = 16];
+  required string message = 4 [(nanopb).max_size = 256];
```
kylestach, line 117: This max size and the max count (16) listed below are up for discussion. These values make the in-memory struct huge (the wire version will cut out space though).

PR #132, opened 2021-09-09
Chipper fix fixed

bperez40:
use_chip isn't defined before line 251. Either needs to be defined earlier or this line needs to be in the execute_cmd function.


Arvind-Srinivasan (approved):
Good

PR #133, opened 2021-09-13
Firmware M1 support

Arvind-Srinivasan:
- Patching install process for M1


bperez40 (approved):
Arvind is the MVP for mac users around the world

PR #134, opened 2021-09-26
Cleanup time

Arvind-Srinivasan:
Our build system might be better now 

shoutout to @ashbhan for helping test this :)


ashbhan (approved):
Works on 20.04

PR #148, opened 2022-02-19
Readme

Arvind-Srinivasan:
Merge the PR in mTrain https://github.com/RoboJackets/mtrain-firmware/pull/62 then look at this PR and merge

PR #156, opened 2022-05-12
Monorepo + fixes

kasohrab:
# Motivation
Making to changes to firmware is a hassle when that change relies upon both the mTrain and firmware repo. To reduce that friction, the best design choice is to merge the two together. This pr makes firmware a monorepo (other than fshare).
# Description
Title. Looks like a lot lol. It seems like it worked since the mTrain git history was preserved.
The +fixes part of the title is me branching off arvind's mtrain and firmware branches (so merging this pr will close those as well, see #148). 
# Other fixes

- unified doxygen setup. 
- set up main.cpp to start tracking module meta data again (since we had a hard time getting FreeRTOS trace to work), will probably actually be made useful in a later pr
- delete unused tests
- lots of cleanup

Later prs will also hopefully add more tests and cleanup the codebases a bit more (and make the doxygen cleaner too). Right now the doxygen actually has everything on it which is good enough for me (before large parts of the codebase were not being tracked by doxygen).

Once this is merged, the mTrain repo will be archived.


Arvind-Srinivasan (changes requested):
I'll come back and look at approving this once we can confirm robots run in shop with the main build.

control/include/drivers/AVR910.hpp
```diff
@@ -38,10 +38,7 @@
 
 #pragma once
 
-#include "mtrain.hpp"
-#include "SPI.hpp"
-#include "DigitalOut.hpp"
-#include "LockedStruct.hpp"
+#include "robocup.hpp"
 #include <memory>
 #include <cstdio>
```
Arvind-Srinivasan, line 43: Stuff like this should be in our standard robocup.hpp file

control/include/drivers/ICM20948.hpp
```diff
@@ -1,10 +1,9 @@
 #pragma once
 
-#include "mtrain.hpp"
-#include "SPI.hpp"
-#include "LockedStruct.hpp"
+#include "robocup.hpp"
 #include <memory>
 #include <vector>
+#include <cmath>
```
Arvind-Srinivasan, line 6: Same with these

control/include/drivers/ISM43340.hpp
```diff
@@ -1,12 +1,15 @@
...
 #include "GenericRadio.hpp"
-#include "pin_defs.h"
 #include <string>
 #include <vector>
 #include <memory>
+#include <cstring>
+
+#include "FreeRTOS.h"
+#include "task.h"
+#include <cstdio>
```
Arvind-Srinivasan, line 12: Move to robocup.hpp the standard lib includes

control/include/robocup/modules/MotionControlModule.hpp
```diff
@@ -1,14 +1,15 @@
 #pragma once
 
 #include "modules/GenericModule.hpp"
-#include "MicroPackets.hpp"
+#include "rc-fshare/robot_model.hpp"
+#include <math.h>
+#include <algorithm>
```
Arvind-Srinivasan, line 6: This has a weird include formatting

control/mtrain/.gitignore
```diff
@@ -0,0 +1,44 @@
+*.kdev4
```
Arvind-Srinivasan, line 1: This whole file shouldn't exist

control/mtrain/API/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+aux_source_directory(Src/ API_CPP_SRC)
```
Arvind-Srinivasan, line 1: The "cpp" folder should be deprecated with instead just an API_SRC folder

control/mtrain/doc/DoxygenLayout.xml
```diff
@@ -0,0 +1,194 @@
+<doxygenlayout version="1.0">
```
Arvind-Srinivasan, line 1: Maybe test to see how we can integrate this into the larger robocup firmware doxygen setup

Arvind-Srinivasan (changes requested):
Before we can merge, a) bunch of minor comments, b) cleaning up the file structure, c) the doxygen docs building should work on the whole, and d) the includes should be cleaned up.

First and foremost, every single test from the mtrain level up should work and imo, we should write even more tests to use as regression benchmarks. Would be good to know every API works for example using control. We should also clean up the tests in control level to have only the most recent hardware tests (I.E. get rid of the ones for an IMU we no longer have)

kasohrab:
being done in #157 instead, but the description I will leave here.

Arvind-Srinivasan:
@kasohrab are we reopening this now 156 is working?

kasohrab:
> @kasohrab are we reopening this now #157 is working?

so on this branch radio wouldn't initialize. I'm not sure what of your changes would cause that and it's hard to debug currently. Also, I'd like to start moving to fix scheduling stuff as soon as I can. 

I was thinking to slowly add some of your tweaks over time until something breaks. But if you know what could have caused that then we could reopen this?

PR #158, opened 2022-06-01
Fix problems

kasohrab:
apparently everything is broken. 
And this pr mostly fixes that. There are some outstanding issues that we can fix this year, the goal of this pr is just to make master usable again.

- ~~Prevents radio from being preempted while receiving and thus always returning no data available.
- Acquires locks in a consistent order (the rtos locks will timeout after a while so that's not the main deadlocking issue; it needed to be fixed though).
- pretty sure there's still a radio state machine bug. We can fix that by adding a timeout to those infinite while loops?
- Add motor test (see motor test hpp cause I was too lazy to create a non-rtos version)
- create new radio test (this doesn't require the rtos, only radio module, mTrain, and a configured computer are needed). Use corresponding python script that Will made to check if both send and receive are working at the driver level.
- Run radio more often and **do not do both send and recieve in the same entry** to keep the RadioModule entry short. 
- many other small improvements (like combining motion control and imu modules)

### DIFF
--- a/control/include/robocup/LockedStruct.hpp
+++ b/control/include/robocup/LockedStruct.hpp
@@ -72,9 +72,7 @@ public:
      *      lock method was called from
      * @return a lock on this struct.
      */
-    Lock lock(bool* first_lock = nullptr) {
-        return Lock(this, first_lock);
-    }
+    Lock lock(bool* first_lock = nullptr) { return Lock(this, first_lock); }
 
     /**
      * Bypass the locking system to directly get a pointer to the underlying
--- a/control/include/robocup/LockedStruct.hpp
+++ b/control/include/robocup/LockedStruct.hpp
@@ -72,7 +72,7 @@ public:
      *      lock method was called from
      * @return a lock on this struct.
      */
-    Lock lock(bool* first_lock = nullptr, const char* debug_string = nullptr) {
+    Lock lock(bool* first_lock = nullptr) {
         return Lock(this, first_lock);
     }
 
@@ -90,7 +90,7 @@ public:
     T* unsafe_value() { return &value; }
 
 private:
-    void acquire_mutex() { xSemaphoreTakeRecursive(mutex, 100); }
+    void acquire_mutex() { xSemaphoreTakeRecursive(mutex, 300); }
 
     void release_mutex() { xSemaphoreGiveRecursive(mutex); }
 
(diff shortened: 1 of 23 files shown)


control/src/robocup/modules/MotionControlModule.cpp
```diff
@@ -33,22 +35,23 @@ MotionControlModule::MotionControlModule(LockedStruct<BatteryVoltage>& batteryVo
 }
 
 void MotionControlModule::entry() {
```
kasohrab, line 37: one lock at a time

control/src/robocup/modules/RadioModule.cpp
```diff
@@ -49,6 +51,14 @@ void RadioModule::start() {
...
 
 void RadioModule::entry() {
+    if constexpr (run_motor_test) {
+        fakeEntry();
+    } else {
+        realEntry();
+    }
+}
+
+void RadioModule::realEntry() {
```
kasohrab, line 61: One lock at a time

control/include/robocup/modules/RadioModule.hpp
```diff
@@ -24,7 +24,7 @@ class RadioModule : public GenericModule {
     /**
      * Priority used by RTOS
      */
-    static constexpr int kPriority = 3;
```
kasohrab, line 27: Revert this

control/include/robocup/LockedStruct.hpp
```diff
@@ -96,10 +96,12 @@ struct LockedStruct {
 private:
     void acquire_mutex() {
         xSemaphoreTakeRecursive(mutex, 100);
```
kasohrab, line 98: https://www.freertos.org/a00122.html
Increase timeout dramatically

control/src/robocup/modules/RadioModule.cpp
```diff
@@ -59,36 +69,68 @@ void RadioModule::entry() {
         battery = batteryVoltage.lock().value();
```
kasohrab, line 69: acquire in consistent order

control/src/robocup/modules/MotionControlModule.cpp
```diff
@@ -1,10 +1,12 @@
 #include "modules/MotionControlModule.hpp"
+
+#include <algorithm>
+
+#include "DigitalOut.hpp"
+#include "MicroPackets.hpp"
 #include "mtrain.hpp"
+
 #include "rc-fshare/robot_model.hpp"
-#include <math.h>
```
kasohrab, line 4: Replaced with <cmath> as it is preferable.

kasohrab:
see #164

